PR #39, opened 2020-12-19
Fix use-after-free of xc. No functional change is intended.

yTakatsukasa:
Address sanitizer told me that `xc->xc_parent` is referred after `free(xc)`.


gtkwave:
That's a good find.  Thanks for generating a patch for this.

PR #57, opened 2021-04-25
Update GHW sources

umarcor:
Some changes were done in GHDL's repo related to C sources for reading GHW files. Mainly, `ghwlib` was renamed to `libghw`, since it is now built as a shared library. In this PR, `ghwlib.*` files are renamed to `libghw.*`. By the way, their content is updated.

Compilation is successful, but I don't whether some specific test exists for ensuring that this update is correct. Specially, whether `ghw.c` needs to be updated because of https://github.com/ghdl/ghdl/commit/ce49befde5e95066ee34469e043a48d94423c53b. I think it's not necessary. @tgingold, can you please confirm?


tgingold:
I don't think so.

PR #71, opened 2021-08-22
Improve drawing of filled waveforms

rfuest:
This PR improves the drawing of filled waveforms. The current code doesn't correctly align the filled rectangle with the pixel grid, which causes some blurring at the edges. This is especially noticeable in black and white mode.

Before (UI scaling: 1, scaled 4x in GIMP)
grafik

After:
grafik

I've also changed the line cap to square, which removes some gaps in the display and results in a crisper display:

Before (UI scaling: 3):
grafik
After:
grafik


gtkwave (approved):

gtkwave:
Looks really good.  Thanks for finding that rendering problem.

PR #72, opened 2021-08-23
Use GtkSearchEntry in SST

rfuest:
This PR replaces the regular GtkEntry in the SST sidebar with a GtkSearchEntry. The search entry contains a search icon, which makes the "Filter:" label redundant. By removing the label and the frame around the buttons the UI looks cleaner.

If you think that PRs like this are useful I would like to continue to tweak the UI/UX in further PRs. But this is made harder by having to maintain the backward compatibility with GTK 2. Do you plan to remove the GTK 2 support from "gtkwave3-gtk3" at some point?

There are two functions which build a similar SST widget: `treebox` and `treeboxframe`. Is the separate `treebox` window still accessible in the GTK3 version? I've only modified `treeboxframe` in this PR.

Before:
Bildschirmfoto vom 2021-08-23 19-16-10

After:
Bildschirmfoto vom 2021-08-23 19-16-53

### DIFF
--- a/gtkwave3-gtk3/src/main.c
+++ b/gtkwave3-gtk3/src/main.c
@@ -314,7 +314,7 @@ if(!GLOBALS->disable_menus)
 			gtk_widget_show(pan_dn);
 			gtk_tooltips_set_tip_2(pan_dn, "Show toolbar");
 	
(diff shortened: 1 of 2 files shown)


gtkwave:
Looks good, thanks.  I'll eventually remove gtk2 support...probably when I start working on a gtk3->4 version.

PR #104, opened 2021-12-18
ci: eine/tip was merged into pyTooling/Actions/releaser

umarcor:
This PR updates the Action used in CI job 'nightly'. By the way, the `workflow_dispatch` event is added, in order to allow manually triggering builds on forks.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,7 @@ on:
   push:
   schedule:
     - cron: '0 0 * * *'
+  workflow_dispatch:
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


rfuest:
Is there a way to update the release date when a new nightly release is uploaded? Or, if this isn't possible, to add the build date to the title or message for the release? At the moment there is no way for the user to know when the nightly build was last update (except for the commit message) and they might assume that it is from 2020.

umarcor:
@rfuest, unfortunately there is not. In order to update the date, the release needs to be removed and created again. That would notify all the watchers explicitly about it, which can be really annoying since this 'nightly' job is executed for each CI run. Hence, the Action updates the assets and force-pushes the tag to point to the latest commit. That updates the title and the body of the 'nightly' release. Overall, everything is correct except the date.

See eine/tip#111.

rfuest:
OK, thanks for the explanation.

umarcor:
I rebased this PR on top of master, and I fixed the warnings due actions needing an update.

umarcor:
@gtkwave @rfuest can we please have this merged?

PR #120, opened 2022-04-25
Remove colors from signal list values

TheZoq2:
This PR removes colors from the signal window to display them as they are displayed in the waveform window.

I'm not a huge fan of this code, modifying the string in place seems kind of ugly to me, but it seems to be what is done in other areas of the code base too.

Here is a screenshot of the result.
image
Previously, the second line would have been `p_read_pos_n151=?DarkOrange?(UNDEF,UNDEF,UNDEF)`

### DIFF
--- a/gtkwave3-gtk3/src/signal_list.c
+++ b/gtkwave3-gtk3/src/signal_list.c
@@ -209,7 +209,37 @@ static void render_signal(cairo_t *cr, Trptr t, int width, int text_dx) {
     if (HasWave(t) || bv) {
         if (t->asciivalue && !(t->flags & TR_EXCLUDE)) {
             int text_x = GLOBALS->max_signal_name_pixel_width + 6 + text_dx;
-            XXX_font_engine_draw_string(cr, GLOBALS->signalfont, &text_color, text_x, text_y, t->asciivalue);
+
+            char* text_no_color = t->asciivalue;
+
+            int has_color = 0;
+            // Look for the first character after the = sign
+            if(*(t->asciivalue + 1) == '?') {
+                // Look for the trailing ?
+                char *srch_for_color = strchr(t->asciivalue+2, '?');
+                // If we found ?
+                if(srch_for_color) {
+                    // Try to see if this is a valid color
+                    *srch_for_color = '\0';
+                    wave_rgb_t cb = XXX_get_gc_from_name(t->asciivalue+2);
+                    if(cb.a != 0.0) { // Found a replacement
+                        // srch_for_color will now point to the end of the color. Set that as
+                        // our actual pointer
+                        text_no_color = srch_for_color;
+                        has_color = 1;
+                    }
+                    *srch_for_color = '?';
+               }
+            }
+
+            // If we had a color, the first character of the printed string
+            // will be '?' rather than '='. Store the original first character,
+            // replace it, draw the string then restore it to its original form
+            char first_char = *text_no_color;
+            *text_no_color = '=';
+
+            XXX_font_engine_draw_string(cr, GLOBALS->signalfont, &text_color, text_x, text_y, text_no_color);
+            *text_no_color = first_char;
         }
     }
 }


rfuest:
In my branch of gtkwave I've started to use more of the string functions provided by glib. These can make the code a lot more readable IMO. Here is my implementation of color parsing:
```c
static void parse_color(const char *text, wave_rgb_t **color_out, char **text_out)
{
    *text_out = NULL;
    *color_out = NULL;

    if (g_str_has_prefix(text, "?")) {
        gchar **tokens = g_strsplit(text, "?", 3);
        if (tokens[0] != NULL && tokens[1] != NULL && tokens[2] != NULL) {
            wave_rgb_t color = XXX_get_gc_from_name(tokens[1]);

            if (color.a != 0.0) {
                *text_out = g_strdup(tokens[2]);

                *color_out = g_new(wave_rgb_t, 1);
                **color_out = color;
            }
        }
        g_strfreev(tokens);
    }

    if (*text_out == NULL) {
        *text_out = g_strdup(text);
    }
}
```

gtkwave:
That looks good...there are a lot more useful convenience functions in glib now than there was back in 1998 (!) when I started working on the viewer and changes like that definitely help with readability.

-Tony

    On Monday, April 25, 2022, 10:23:50 AM EDT, Ralf Fuest ***@***.***> wrote:  
 
 


In my branch of gtkwave I've started to use more of the string functions provided by glib. These can make the code a lot more readable IMO. Here is my implementation of color parsing:
static void parse_color(const char *text, wave_rgb_t **color_out, char **text_out)
{
    *text_out = NULL;
    *color_out = NULL;

    if (g_str_has_prefix(text, "?")) {
        gchar **tokens = g_strsplit(text, "?", 3);
        if (tokens[0] != NULL && tokens[1] != NULL && tokens[2] != NULL) {
            wave_rgb_t color = XXX_get_gc_from_name(tokens[1]);

            if (color.a != 0.0) {
                *text_out = g_strdup(tokens[2]);

                *color_out = g_new(wave_rgb_t, 1);
                **color_out = color;
            }
        }
        g_strfreev(tokens);
    }

    if (*text_out == NULL) {
        *text_out = g_strdup(text);
    }
}

gtkwave:
What is generating these color strings such as the "?Dark Orange?" value?I've never seen this problem with transactional traces before.  What specifically are you doing?  Something seems wrong here.

-Tony


    On Monday, April 25, 2022, 07:50:58 AM EDT, Frans Skarman ***@***.***> wrote:  
 
 
This PR removes colors from the signal window to display them as they are displayed in the waveform window.

I'm not a huge fan of this code, modifying the string in place seems kind of ugly to me, but it seems to be what is done in other areas of the code base too.

Here is a screenshot of the result.

Previously, the second line would have been p_read_pos_n151=?DarkOrange?(UNDEF,UNDEF,UNDEF)

You can view, comment on, or merge this pull request online at:

  https://github.com/gtkwave/gtkwave/pull/120

Commit Summary
   
   - 07a5994 Remove colors from signal list values

File Changes 
 
(1 file)
   
   -  M gtkwave3-gtk3/src/signal_list.c (32) 

Patch Links:
   
   - https://github.com/gtkwave/gtkwave/pull/120.patch
   - https://github.com/gtkwave/gtkwave/pull/120.diff

TheZoq2:
Yeah, the glib solution looks much neater.

The colors come from strings in my VCD file which are generated by another tool. The reason I want them is that I'm writing a custom HDL with a much more expressive type system than verilog, and reading raw binary data is pretty difficult. I therefore have a tool which takes a VCD file and outputs a "parsed" VCD file as strings. The colors come directly from the those strings, i.e. I emit `"?DarkOrange"(UNDEF, UNDEF, UNDEF)` into the vcd.

gktwave seems to handle that fine, apart from this problem in the signal window.

I also looked at doing both of these tasks with translation process filters, but that proved to be fairly difficult. I'd want the filters to be applied automatically by default, and for the type conversion the filter process needs to know both the names and values of the signals.

I also tried to do just the coloring with those filters, but it seems like gtkwave ignores the filter output for string values.

Maybe I could do it with transaction filters (?) but that would require a bit more effort from my side

gtkwave:
If you can, email me the vcd and save file.  There should be an easier way to do this.

-Tony

    On Monday, April 25, 2022, 11:29:56 AM EDT, Frans Skarman ***@***.***> wrote:  
 
 


Yeah, the glib solution looks much neater.

The colors come from strings in my VCD file which are generated by another tool. The reason I want them is that I'm writing a custom HDL with a much more expressive type system than verilog, and reading raw binary data is pretty difficult. I therefore have a tool which takes a VCD file and outputs a "parsed" VCD file as strings. The colors come directly from the those strings, i.e. I emit "?DarkOrange"(UNDEF, UNDEF, UNDEF) into the vcd.

gktwave seems to handle that fine, apart from this problem in the signal window.

I also looked at doing both of these tasks with translation process filters, but that proved to be fairly difficult. I'd want the filters to be applied automatically by default, and for the type conversion the filter process needs to know both the names and values of the signals.

I also tried to do just the coloring with those filters, but it seems like gtkwave ignores the filter output for string values.

Maybe I could do it with transaction filters (?) but that would require a bit more effort from my side

PR #158, opened 2022-10-09
Improve bundle directory detection.

DanielO:
Now works no matter what the current directory is allowing the binary to be run out of the application from the command line.

Note that I have only tested this by editing the script in the prebuilt application downloaded from https://sourceforge.net/projects/gtkwave/files/gtkwave-3.3.107-osx-app/ as I don't have the app build environment setup.


gtkwave:
Thanks for the patch.

PR #161, opened 2022-10-25
Disable stdio buffering when accessing FST files on NetBSD

lparkes:
This is the same as OpenBSD and FreeBSD.

I tested the code in gtkwave3, gtkwave3-gtk3 and gtkwave4. I could only test that my change compiled in gtkwave4, but I was able to run the other versions of gtkwave and confirm that I can open FST files on NetBSD.


gtkwave:
Thanks for the pull request.
-Tony

PR #188, opened 2023-02-25
MSYS2: CI packaging fixes

dlmiles:
Run pacman -U --noconfirm mingw-*gtkwave*.pkg.tar.zst
  pacman -U --noconfirm mingw-*gtkwave*.pkg.tar.zst
  mv mingw-*gtkwave*.pkg.tar.zst ../../
  ../standalone_pkg.sh

../standalone_pkg.sh
  cp: cannot stat '/mingw64/bin/libffi-7.dll': No such file or directory

### DIFF
--- a/MSYS2/dll.inc
+++ b/MSYS2/dll.inc
@@ -6,7 +6,7 @@ libcairo-2.dll
 libcairo-gobject-2.dll
 libdatrie-1.dll
 libexpat-1.dll
-libffi-7.dll
+libffi-8.dll
 libfontconfig-1.dll
 libfreetype-6.dll
 libfribidi-0.dll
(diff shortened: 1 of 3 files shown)


MSYS2/gtk3dll.inc
```diff
@@ -1,3 +1,4 @@
+libcairo-gobject-2.dll
 libepoxy-0.dll
 libgdk-3-0.dll
 libgtk-3-0.dll
```
umarcor, line 4: Maybe add a newline at the end of this file.
It was missing already, not changed by this PR.

umarcor (approved):
LGTM!

MSYS2/gtk3dll.inc
dlmiles: Understand that is my preference too (all txt file have EOL as last character sequence) but as a contributor I assume the project policy is as it is already, so it was left.

umarcor: Absolutely. Just pointed out because I saw it there. This PR is good to merge as-is.

dlmiles:
Force push minor change to remove the entry, MSYS2/gtk3dll.inc libcairo-gobject-2.dll

Because it is already correctly present in MSYS/dll.inc and becomes duplicated/redundant in MSYS2/gtk3dll.inc from my original patch in this PR.

Like the previous version this also passes CI now.

PR #190, opened 2023-03-12
Define FST_DO_MISALIGNED_OPS on AArch64

nickg:
AArch64 handles misaligned loads/stores in hardware.  In fact fstGetUint32 gets compiled into a plain load on -O2:

```
fstGetUint32:
        ldr     w0, [x0]
        ret
```

### DIFF
--- a/gtkwave3-gtk3/src/helpers/fst/fstapi.c
+++ b/gtkwave3-gtk3/src/helpers/fst/fstapi.c
@@ -131,7 +131,7 @@ void **JenkinsIns(void *base_i, const unsigned char *mem, uint32_t length, uint3
 #define FST_GZIO_LEN                    (32768)
 #define FST_HDR_FOURPACK_DUO_SIZE       (4*1024*1024)
 
-#if defined(__i386__) || defined(__x86_64__) || defined(_AIX)
+#if defined(__i386__) || defined(__x86_64__) || defined(_AIX) || defined(__aarch64__)
 #define FST_DO_MISALIGNED_OPS
 #endif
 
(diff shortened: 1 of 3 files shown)


rfuest:
Maybe it would also be possible to get rid of FST_DO_MISALIGNED_OPS instead. The compiler will output the same code, if it is written with `memcpy` and optimization is enabled: https://godbolt.org/z/naoc47Woq

nickg:
> Maybe it would also be possible to get rid of FST_DO_MISALIGNED_OPS instead. The compiler will output the same code, if it is written with `memcpy` and optimization is enabled: https://godbolt.org/z/naoc47Woq

I'd be tempted to do it the other way round and get rid of `fstGetUint32` completely: what common CPU that users are likely to run GtkWave on _doesn't_ support unaligned accesses? Even 32-bit Arm supported it since armv7-a.

rfuest:
> I'd be tempted to do it the other way round and get rid of `fstGetUint32` completely: what common CPU that users are likely to run GtkWave on _doesn't_ support unaligned accesses? Even 32-bit Arm supported it since armv7-a.

RISC-V isn't very common yet, but I think it is an example of an architecture that doesn't support unaligned access.

But IMO it doesn't matter how common unaligned access is. Without `fstGetUint32` the code for the conversion of bytes to `uint32_t` would have to be duplicated multiple times, which makes the code less readable and maintainable. Using `memcpy` is also the safest option to not introduce any issues with the strict aliasing rule.

nickg:
AFAIK there is no aliasing issue when accessing through `unsigned char *` as a pointer to a character type may legally alias anything. But I agree `memcpy` would be an improvement over what we have now. Or perhaps use a union which compiles to the same thing: https://godbolt.org/z/Pfs6P7o9K

rfuest:
Sure, a union would also work. I prefer the `memcpy` version, because it doesn't contain any magic values, but any version would be OK with me.

The rest of the FST API code seems to still be C89 compatible, which the initializer in the `union` code is not. I don't known if Tony wants to keep compatibility with these ancient C compilers for the FST API. Other parts of GTKWave already use C99 features.

gtkwave:
If the FST API code uses C99, sure, that's fine.
As far as memcpy goes, I have no preference as long as the code is functionally correct.

gtkwave:
Also...merged, thanks.

PR #191, opened 2023-03-12
Use MapViewOfFileEx for mmap on Windows

nickg:
This was reported by a user in nickg/nvc#596.  The FST writer emulates mmap on Windows by copying from the file to a temporary buffer and then flushing that buffer to disk in fstDestroyMmaps.  The problem here is that while `xc->curval_mem` is flushed, `xc->valpos_mem` is not.  One fix would be to duplicate the flushing logic for valpos_mem but it seems better to just use the native Windows MapViewOfFileEx API.  Cygwin has its own mmap emulation layer so the existing POSIX code should work fine there.

I've been using this patch for a while with no reported issues, and I did a full build of GtkWave on MSYS2 and checked it could load FST files created with it.

### DIFF
--- a/gtkwave3-gtk3/src/helpers/fst/fstapi.c
+++ b/gtkwave3-gtk3/src/helpers/fst/fstapi.c
@@ -58,6 +58,7 @@
 #endif
 
 #ifdef __MINGW32__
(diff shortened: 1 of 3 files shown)


gtkwave:
Merged, thanks.

PR #205, opened 2023-06-04
Add %F to exec key in desktop file

aelmahmoudy:


rfuest:
I don't think `%F` is the correct field in this instance. GTKWave can't open multiple dumpfiles in one command line. The second and third optional file arguments in the command line can be used to specify a savefile and rcfile. But I don't think that relying on the order of files to passed as `%F` is a good idea. `%f` would be a better choice to make sure only one file is passed.

PR #210, opened 2023-06-18
Space as separator for large time values

jotego:
I'm dealing with a file that contains several seconds of simulation at a MHz clock and reading the time values in GTKWave goes beyond my mental abilities. 

I have modified the `reformat_time` to separete by thousands when there is no `scale_to_time_dimension` set. It looks like this:

image

when you zoom in:

image

I think this probably needs some other changes in the code (maybe in the from/to buttons). If you give me some orientation about what else needs to be done to get this PR accepted, I'll do it. Please bear in mind that I know C but I do not know GTK.

### DIFF
--- a/gtkwave3-gtk3/src/currenttime.c
+++ b/gtkwave3-gtk3/src/currenttime.c
@@ -326,7 +326,7 @@ while( aux[s] )
 	buf[d++]=aux[s++];
 	if(--c==0 && aux[s] )
 		{
-		buf[d++]=',';
+		buf[d++]=' ';
 		c=3;
 		}
 	}
--- a/gtkwave3-gtk3/src/currenttime.c
+++ b/gtkwave3-gtk3/src/currenttime.c
@@ -21,7 +21,6 @@ static char *marker_label_text ="Marker Time";
 static char *maxtime_label_text_hpos="Max";
 static char *marker_label_text_hpos ="Marker";
 
-
 #ifdef WAVE_ALLOW_GTK3_HEADER_BAR
 static void update_labels_to_header_bar(void)
 {
@@ -29,7 +28,7 @@ if(GLOBALS->header_bar)
 	{
 	char buf[1024];
 
-	sprintf(buf, "%s" ": " "%s" "  |  " "%s" ": " "%s",     
+	sprintf(buf, "%s" ": " "%s" "  |  " "%s" ": " "%s",
 	        gtk_label_get_text(GTK_LABEL(GLOBALS->max_or_marker_label_currenttime_c_1)),
 	        gtk_label_get_text(GTK_LABEL(GLOBALS->maxtimewid_currenttime_c_1)),
 	        gtk_label_get_text(GTK_LABEL(GLOBALS->base_or_curtime_label_currenttime_c_1)),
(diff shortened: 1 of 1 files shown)


rfuest:
I don't think this should be enabled by default. Number formats are different in different parts of the world and as a german I would interpret the number in the first example as `12.7 * 10^-3` and not `12700`. According to this Wikipedia article https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping a thin space is the closest thing to a universal standard for the grouping of digits.

jotego:
I think that all engineers should just use commas in this case and get rid of the French influence here. Engineering applications seem to ignore locale information for a good reason.

Anyway, this is your application, if you want spaces, we can make it spaces.

rfuest:
> Anyway, this is your application, if you want spaces, we can make it spaces.

Not my application, just my opinion.

jotego:
Sorry for my assumption. My point is that professional engineering applications don't use locale for formatting numbers and that helps sharing information and documents across multi-cultural teams.

rfuest:
> My point is that professional engineering applications don't use locale for formatting numbers and that helps sharing information and documents across multi-cultural teams.

And that's why I suggested a thin space as a default group delimiter. It's unambiguous and won't be interpreted as a decimal point by anyone. For users who prefer a comma there could still be an option to change the delimiter in the configuration file.

gtkwave:
Some form of space is fine, no different than "_" used as a visual separator for groups of digits in Verilog.  In another world, I probably would've handled excessively large time values with a popup window, or have some kind of math function that can convert them to something more useful such as a cycle count, etc.  Many sim testbenches dump cycle counts for that reason.

-Tony


   On Sunday, June 18, 2023 at 12:03:48 PM EDT, Ralf Fuest ***@***.***> wrote:  
 
 



My point is that professional engineering applications don't use locale for formatting numbers and that helps sharing information and documents across multi-cultural teams.


And that's why I suggested a thin space as a default group delimiter. It's unambiguous and won't be interpreted as a decimal point by anyone. For users who prefer a comma there could still be an option to change the delimiter in the configuration file.

jotego:
I have replaced the comma by a space.

By the way, in my particular case, the time value is much more convenient than the number of cycles.

rfuest:
Could we please us a thin space instead? A regular space is too wide and makes the value harder to read.

jotego:
> Could we please us a thin space instead? A regular space is too wide and makes the value harder to read.

I do not know how to encode a thin space in an ASCII string.

rfuest:
All strings in GTK are UTF-8.

jotego:
So what should go instead of `buf[d++]=' ';` ?

rfuest:
There might be easier ways, but this should work:

```c
/* Append UTF-8 encoded small space U+2009 */
buf[d++] = '\xE2';
buf[d++] = '\x80';
buf[d++] = '\x89';
```

jotego:
> There might be easier ways, but this should work:
> 
> ```c
> /* Append UTF-8 encoded small space U+2009 */
> buf[d++] = '\xE2';
> buf[d++] = '\x80';
> buf[d++] = '\x89';
> ```

Thank you, I have added it to the PR

rfuest:
I did just try it and noticed a few issues:
- The group separators shouldn't be inserted in editable values, like the from/to entries.
  Because these entries are changed in multiple places in the code I would suggest that you move the new `reformat_time` variant to a new method, like `reformat_time_with_grouping`, and explicitly pick the places where values are only displayed.
- The grouping isn't applied if scale to dimension is used.

jotego:
I apologize but this task goes over my availability. I have tried hiring a team that supports Verilator to work on GTKWave but they were not willing to engage on this project. I am closing this for now.

Issue #212, opened 2023-06-22
gtkwave crashes after trying to reload a deleted trace


LiamSkirrow:
My simulation flow involves deleting and recreating the directory containing my waveforms from scratch. It seems that gtkwave doesn't like this approach as when I reload the waveform (ctrl+shift+R) it just hangs for a while and I have to force quit the program.

I'm running gtkwave v3.3.61 on Ubuntu 20.04.6 LTS. I'm happy to take care of this in a PR, just let me know roughly where I should be looking in the source code :+1:

rfuest:
Did you check if the issue still exists in newer versions of GTKWave? 3.3.61 is 9 years old. You could, for example, try to run the Flatpak version.

LiamSkirrow:
Fair enough, I'll run it on a newer version and will report back. Thanks.

rfuest:
The FST error has been fixed in release 3.3.115, but apparently the Flatpak version hasn't been updated yet.

LiamSkirrow:
Just to add another datapoint, I tried running gtkwave v3.3.103 and I still have to force close the program when reloading on a deleted trace. 

I don't have immediate access to a more up to date version currently but happy to keep investigating. Where can I actually download the most up to date version of gtkwave3? Shall I build source from the `gtkwave3` dir in this repo to get the latest version? Thanks.

LiamSkirrow:
Ok I built from source (gtkwave3) to run v3.4.0 (Copyright message suggests this dates from 2022 so is probably the most recent version) and can confirm the bug is still persisting, exactly the same with the other versions. 

I'll get working on a fix for this and will submit a PR soon enough :sunglasses:

rfuest:
The `gtkwave3-gtk3` contains the current GTK 3 version. Be sure to use the `--enable-gtk3` flag when you call `./configure`. There are some build instructions in the readme file at https://github.com/gtkwave/gtkwave/tree/master/gtkwave3-gtk3.

rfuest:
I wasn't able to reproduce the error. Could you provide a set of files and a step by step instruction which triggers the bug?

LiamSkirrow:
I was able to reproduce the error by building `gtkwave3-gtk3` following the instructions in the readme.

To reproduce:
```
- cd ./dir-containing-trace/
- gtkwave test.fst
- mkdir tmp && mv test.fst ./tmp    # move the trace to a different area, alternatively just delete it
- [reload waveform -> ctrl+shift+R] 
[...]

PR #213, opened 2023-06-23
Call gtkwave_main_iteration during reload retries

rfuest:
Not a great fix, but it does at least keep Gnome shell from wanting to kill the process.


LiamSkirrow:
Link to issue #212

Issue #219, opened 2023-07-20
Creating releases from the GitHub version


rfuest:
At the moment the only available packaged versions of the GitHub version of GTKWave are the nightly windows builds. I think it would be good to also release the GitHub version to users. A GTKWave 4 release is still a long way out and IMO it's a good idea to fix some issues with GTKWave 3 in the meantime.

One example, that came up recently, is the Wayland support in GTKWave 3. I've recently switched to Wayland and would like to get GTKWave 3 working properly working on Wayland. But doing this correctly will require extensive code changes that won't be able to be included in the SourceForge version immediately. And I don't want to spend any time on this if no user will be able to benefit from this work.

My suggestion to solve this problem would be to release two versions of GTKWave. One regular version from the code that is on GitHub and a LTS release which is based on the SourceForge code. At the moment most users will not be aware that two versions exist and adding a LTS suffix should make the distinction between the versions clear.

My plan to implement this change:
- Move all development to GitHub
  The three subdirectories for the different GTKWave versions will be replaced by three branches:
  - `master`: Contains the content of the `gtkwave3-gtk3` directory. Any further development of GTKWave 3 will happen on this branch.
  - `lts`: Contains the SourceForge version of GTKWave. Only important fixes will be cherry picked from the `master` branch.
  - `gtkwave4`: Contains the development version of GTKWave 4.  
  - The `gtkwave3` directory will be deleted, because it's outdated anyway.
- The `master` branch will be released as "GTKWave 3.4.xxx".
- The stable version, which had been maintained on SourceForge, will be renamed to "GTKWave LTS" and will continue to be released as "3.3.xxx".
- GIT tags and GitHub releases will replace the tar files on SF.

umarcor:
IMHO, changing to the three branch structure would be desirable regardless of making the change to GitHub as the main location. I am also in favour of making GitHub the main location, but I wouldn't like that to be a stopper for the structure reorganisation.

With regard to generating packaged versions for other platforms and/or generating packages for multiple branches/versions, please let me know if you need help setting up CI.

gtkwave:
My suggestion to solve this problem would be to release two versions of GTKWave. One regular version from the code that is on GitHub and a LTS release which is based on the SourceForge code. At the moment most users will not be aware that two versions exist and adding a LTS suffix should make the distinction between the versions clear.
Message ID: ***@***.***>
[...]

Issue #144, opened 2022-08-10
Convert the user guide to Markdown


rfuest:
At the moment the user guide is a ODT file in the repo, which is also available as a PDF file on the SourceForge website. From previous issues on GitHub I suspect that some users don't know the user guide even exists. The link is easy to miss on the website and it is only linked from the Help menu in the macOS version.

My suggestion is to convert the user guide into Markdown and use a static website generator to generate an online version of the user guide. I've create a proof of concept conversion based on [Material for MkDocs](https://squidfunk.github.io/mkdocs-material/), which can be viewed at: http://gtkwave.rfuest.de. The source is available at: https://github.com/rfuest/gtkwave/tree/docs/docs

Advantages of this approach are IMO:
- Markdown is better suited to be handled by GIT than the binary ODT file.
   This will make it easier for others to contribute to the user guide and to review the changes in PRs.
- Setting up automatic deployment to GitHub Pages seems to be easy: https://squidfunk.github.io/mkdocs-material/publishing-your-site/#github-pages
- The proof of concept website is also intended to replace http://gtkwave.sourceforge.net/
  Having only one website, which is automatically generated from the GIT repository, will make maintenance easier. The SF website could then be changed to redirect to the new `gtkwave.github.io`, like we recently discussed in the PR regarding the application ID.
- The generated website automatically supports modern web features like responsive design or syntax highlighting for code examples.

If we decide to use this approach the docs still need some formatting fixes and someone needs to go through the docs and remove some outdated parts, like compiling for GTK-1.2. But the proof of concept should give a good indication of what the website could look like.

gtkwave:
That looks pretty good.  I think it's a good plan of action to take.

-Tony

    On Wednesday, August 10, 2022, 12:11:13 AM EDT, Ralf Fuest ***@***.***> wrote:  
 
 


At the moment the user guide is a ODT file in the repo, which is also available as a PDF file on the SourceForge website. From previous issues on GitHub I suspect that some users don't know the user guide even exists. The link is easy to miss on the website and it is only linked from the Help menu in the macOS version.

My suggestion is to convert the user guide into Markdown and use a static website generator to generate an online version of the user guide. I've create a proof of concept conversion based on Material for MkDocs, which can be viewed at: http://gtkwave.rfuest.de. The source is available at: https://github.com/rfuest/gtkwave/tree/docs/docs

Advantages of this approach are IMO:
   
   - Markdown is better suited to be handled by GIT than the binary ODT file.   
[...]

PR #220, opened 2023-07-21
Make gtkwave3-gtk3 directory the new root

rfuest:
This PR moves the content of the `gtkwave3-gtk3` into the root of the repository, as discussed in #219. The `gtkwave4` directory will later be restored as a separate branch.


README.md
```diff
@@ -1,6 +1,147 @@
 # GTKWave
+
 GTKWave is a fully featured GTK+ based wave viewer for Unix and Win32 which reads LXT, LXT2, VZT, FST, and GHW files as well as standard Verilog VCD/EVCD files and allows their viewing.
 
-This repository contains two versions of GTKWave:
```
umarcor, line 4: Since this will become the main branch, what people read when they clone the repo or they visit it in the browser, I believe it would be desirable a note in this regard. For example:

"This branch contains the gtkwave3-gtk3 version with support for GTK+ 2 and GTK+ 3. See also branches XXXX and YYYYY."

rfuest: That makes sense. I'll include this change in a future PR, because I also want to update the README a bit more. At the moment I want to at least include a screenshot and a link to the PDF documentation in the README. At a later date this could then be updated to point to the new docs website, which I started in #144 and didn't have time to finish and convert to Sphinx yet.

gtkwave:
I checked the version bump to 3.3.116 into SF and put out the tarfiles in the files area.  That should be the final update to SF.  At some point I'll update the webpage saying that it exists for historical reasons.  I don't know what SF terms of service are as it's been forever since I've read them, but my guess is that sending a link to a current GitHub probably isn't on their users' desired behavior list.

I remember the days when I used to post announces on Freshmeat back when it was a new, sister site to Slashdot.  That was *so* long ago...

-Tony

rfuest:
Do you have any remaining comments on this PR or the `lts` branch? Otherwise I would merge this commit and push the `lts` branch to this repo, after I've added the last SVN commit.

I'm not sure what SF's terms and conditions say, but I've seen multiple projects that link to GitHub. Some even have a GitHub link instead of the regular `Code` button in the tab bar. Perhaps that is done by using the GitHub importer (https://sourceforge.net/p/import_project/github/), but I'm not sure if is possible to use the importer for an existing project to keep the SF page alive as a backup.

gtkwave:
I have no real comments outside of the "version bump to 3.3.116".  

I'll look at SF terms.  My main concern more is along the lines that there are going to be pages all over that link to the SF page and having it go completely black wouldn't be the best thing.  I'll look to see what's doable.  However, if they kill the page, they kill the page: it's their site and I'm not paying for hosting on it.  I got 16 years out of SF which is 15 years more than I ever expected.

-Tony

PR #224, opened 2023-07-29
Remove GTK 2 support and improve drawing performance

rfuest:
This PR removes GTK 2 support and fixes various other problems, see the Changelog for a summary of changes. I'll leave this as a draft pull request for the moment, because the drawing code still needs further testing.


rfuest:
I've just updated the PR to remove the headerbar and to replace the hambuger menu with a traditional menubar.

Another thing I changed is how the marker and cursor times are displayed. The time values were automatically rounded to other units if they ended in zeros, which caused the labels to jump around if you did move the cursor or marker. In the new code the times are displayed right aligned and always with same unit. To improve readability for longer time values the digit grouping are now grouped, like previously suggested in #210.

https://github.com/gtkwave/gtkwave/assets/226123/241946a1-32a0-46de-89e9-8b086158b607

rfuest:
This branch now uses meson instead of autotools. I still need to do more testing on other OSs, but most options from the autotools version should be available. But there are some regressions:
- RPC isn't supported
  The code uses GSettings as a RPC mechanism, which didn't work in my tests. If I understand the history of this feature correctly it has been ported from GConf, which was used as a way to implement a D-Bus RPC. Are there many users of this feature? If we want to keep RPC working I would suggest to implement D-Bus RPC directly and not indirectly via GSettings/GConf.
- Tcl support cannot be compiled using Tcl stubs
  Didn't have time to look into this yet.
- AET2, FSDB and Extload aren't supported
 I don't have a way to test these features.
- GEDIT_PATH isn't set
  Detecting gedit at build isn't very useful for packaged versions of GTKWave. gedit has also been replaced by the new Gnome Text Editor in recent versions of GNOME.

gtkwave:
1) The main reason that gconf/gsettings/whatever RPC is in existence is so I can auto-reload immediately after sim finishes:
    ./simv    gconftool-2 --type string --set /com.geda.gtkwave/0/reload 0

So instead of manually having to click reload, once sim finishes, it'll reload waves for me, which is often when I've run off for a cup of coffee or something.  I don't believe there are any other real usages out there, except for creative periodic use of writesave.  When users complain, "Why doesn't gtkwave periodically autosave?"  Well, actually it does but it's kind of hidden...

Any mechanism that eventually works with respect to triggering off reload asynchronously is fine, but not supporting RPC isn't a dealbreaker as that's easy to add in.


2) Tcl stubs AFAIK was only used by one individual, a contractor named Mr. Yiftach Tzori.  I believe he wrote his own Debussy-like integrated toolset driven by either emacs or vim, and the way he did things used stubs elsewhere so gtkwave followed suit as an option to accommodate him.  If his name ever comes up (this was quite long ago, so perhaps not), it'll likely be related to this.  Most of the interesting Tcl function in gtkwave was from either him or Concept Engineering and I learned a lot about Tcl interpreters from Yiftach.


3) Gedit was an arbitrary choice.  Anything as a text editor works.  As you say with the packaging, unless it's listed as a package dependency when the RPM or whatever is put together, it can't reasonably be expected to exist at runtime.  For various distros, I'd have personal correspondence with those distro maintainers so we were in sync with expectations.

-Tony

rfuest:
1: I'll look into porting GTKWave to use the `GtkApplication` class to handle the lifecyle of the program, which I hadn't plan to include in 3.4 but which will be necessary for GTK 4 anyway. This way all actions will be automatically available via D-Bus. But this will at least need some changes to the massive `main_2` function, which I'm not looking forward to.

3: [g_file_query_default_handler](https://docs.gtk.org/gio/method.File.query_default_handler.html) should make it possible to open the files in the default application which is associated with the filetype. I will remove the dependency on gedit and use that function instead, if the `GTKWAVE_EDITOR` env variable and the `editor` rc variable aren't set.

rfuest:
In order to remove some old pixmaps I changed the version dialog to use the `GtkAboutDialog` widget, that is built into GTK 3. The information in the new dialog is based on the old dialog and the `AUTHORS` file in the root of the project. Please check if the information is correct or if anything needs to be updated.

grafik

grafik

The authors list is truncated in the screenshot, the entire list is visible here: https://github.com/rfuest/gtkwave/blob/d249758c0419265d01b3e9fa0390cb43e23934b4/src/menu.c#L1866-L1876 I did only include authors who directly contributed to GTKWave and not one of it's dependencies.

gtkwave:
The contents of AUTHORS and how it's used to generate that info widget look correct to me.At some point, add yourself to the AUTHORS file given that you're doing a comprehensive refactoring/rewrite.

-Tony

rfuest:
I've just pushed an update to this PR that removes a bunch of rc variables, which should be safe to remove: https://github.com/rfuest/gtkwave/blob/234c19bd6e81ce0aaa7f5c4bccb509ca114f3c19/CHANGELOG.md?plain=1#L30-L40

Most variables weren't used anyway, but two variables will remove some alternate modes:
- `use_standard_clicking` shouldn't be necessary anymore with the refactored signal list widget, because traces are no longer deselected when a drag gesture is detected.
- The primary reason to remove the `alt_wheel_mode` setting was to reduce code complexity. In order to prepare for GTK 4 all widgets will need to be ported to only use gestures. In my opinion, having one well tested control scheme is better than having to maintain two schemes.

gtkwave:
I think use_standard_clicking was legacy from early versions of the the gtk-1.2 version.  The original clicking mechanism was that selecting signals would reverse their current state unless you dragged along the y-axis then the most recent state a signal changed to would drag-propagate.  Back in the early 2000s when I was at IBM, Lotus Notes used such kind of a selection mechanism for emails, and I think nWave might still have something similar.  I caught a lot of criticism for implementing such a "weird" selection mechanism, so after a while, it was changed to reflect what other gtk applications use.  After a while on gtk4, I am sure the self-appointed gtk compliance police who don't even know what an HDL is will come out of the woodwork to make disparaging comments.  :-)

I don't recall using alt_wheel_mode with the value of "off" for a very long time.  What's happened with many of these rc vars is a user would suggest new behavior as the default and the old behavior eventually never saw usage.  I agree it's best to have one well-tested control scheme rather than carry around obsolete ones likely nobody has used in years.
Regarding these rc vars, as the cliche goes, in some cases, "Less is more."

-Tony

rfuest:
> I think use_standard_clicking was legacy from early versions of the the gtk-1.2 version.  The original clicking mechanism was that selecting signals would reverse their current state unless you dragged along the y-axis then the most recent state a signal changed to would drag-propagate.  Back in the early 2000s when I was at IBM, Lotus Notes used such kind of a selection mechanism for emails, and I think nWave might still have something similar.  I caught a lot of criticism for implementing such a "weird" selection mechanism, so after a while, it was changed to reflect what other gtk applications use.  After a while on gtk4, I am sure the self-appointed gtk compliance police who don't even know what an HDL is will come out of the woodwork to make disparaging comments.  :-)

I hope they won't suggest to use the selection method that some GNOME applications use, where you need to switch to a special selection mode to make the applications touch compatible ;). There are modern applications that use a similar method to what you've described, but not for selection. In graphics applications, like Inkscape, it is often possible to change the visibility of multiple layers by toggling the visibility of one and then copy the state to other layers by dragging over them. This works extremely well for that use case, but I don't think I would want to use it for selection.

While we are discussing older GUI patterns, which we've used a long time ago in other software. I can still remember using the help function in Word 6 and Excel 4 in Windows 3.1, which worked similar to GTKWaves help function. Getting a longer description, than would be possible in a tooltip, is a nice feature but I would like to remove this feature anyway. My issue is that opening the help window blocks the actions from executing, which isn't ideal if RPC or scripts call the same actions. As a replacement we could link to the new website, when it is migrated to Sphinx and/or bundle a PDF version of the manual, like the macOS version did.

> I don't recall using alt_wheel_mode with the value of "off" for a very long time.  What's happened with many of these rc vars is a user would suggest new behavior as the default and the old behavior eventually never saw usage.  I agree it's best to have one well-tested control scheme rather than carry around obsolete ones likely nobody has used in years. Regarding these rc vars, as the cliche goes, in some cases, "Less is more."

My goal for the refactoring is to make GTKWave behave like a modern application in it's default configuration, which will make it easier to use for new users. But I hope that long time users will be as vocal as the GTK compliance police ;) to request restoring of removed features if that impacts their workflow.

gtkwave:
Sure, remove the help window function if it makes things run smoother such as during scripting, RPC, etc.  I added that one originally as a quick and dirty "I'm too lazy to look at documentation" feature for menu items, but I don't think it's of too much relevance these days.

-Tony

PR #226, opened 2023-08-05
Change project license in appdata

rfuest:
According to `LICENSE.TXT` and the individual file headers the license is "... either version 2 of the License, or (at your option)
any later version." I think that means that the SPDX identifier in the appdata is incorrect, but I'm not very familiar with the intricacies of open source licensing.

### DIFF
--- a/share/appdata/gtkwave.appdata.xml
+++ b/share/appdata/gtkwave.appdata.xml
@@ -3,7 +3,7 @@
 <component type="desktop">
   <id>io.github.gtkwave.GTKWave</id>
   <metadata_license>CC0-1.0</metadata_license>
-  <project_license>GPL-2.0-only</project_license>
+  <project_license>GPL-2.0-or-later</project_license>
   <name>GTKWave</name>
   <developer_name>Tony Bybell</developer_name>
   <update_contact>bybell@rocketmail.com</update_contact>


gtkwave:
Looks good, thanks.

gtkwave:
They put "or any later version" in the licensing as an automatic GPL version upgrade escape hatch.  I'm fine with your change.

I don't remember what the details were with GPLv3 when it came out that I didn't like versus GPLv2, but that was years ago and it doesn't really matter at this point.  I don't have any religious-level convictions regarding open source licensing--especially on a code base that goes back 25 years, but probably more like 35 if you look at some of the header stuff in analyzer.h that originally came from the wave viewer that shipped with the VLSI tool Magic!

The only real licensing issue in the whole codebase that I'm concerned with is that readers/writers will always have zero restrictions, basically to allow any tools to use them without handstands.  At this point, if someone wants to improve FST by cleaning out the backward compatibility gunk and close sources it, he's going to do that.  Concepts are more important in this realm of compression algorithms than implementations, anyway.  I didn't connect all the dots for FST until about 2009, and that was after I have been doing trial and error experimentation with other formats for almost a decade and reverse engineering commercial and IBM EDA internal formats, talking to IBM developers about "what really works," etc.  I've put concept material enough out there where someone could look at that work and have a way better starting point than I did.

-Tony

umarcor:
> the header stuff in analyzer.h that originally came from the wave viewer that shipped with the VLSI tool Magic

Do you mean http://opencircuitdesign.com/magic/ ? (an earlier version of it)

gtkwave:
"Do you mean http://opencircuitdesign.com/magic/ ? (an earlier version of it)"

Yes, specifically this part of the Xt-based "analyzer" part of irsim in that tool suite:http://opencircuitdesign.com/irsim/

515:/tmp/p/irsim-9.6.4/src/analyzer> ls -la ana.h 
-rw-r--r--. 1 bybell bybell 4907 Feb 18  2003 ana.h524:/tmp/p/irsim-9.6.4/src/analyzer> ls -la analyzer.c 
-rw-r--r--. 1 bybell bybell 6919 Nov 14  2002 analyzer.c

You can see where some of the data structs and functions got their inspiration or were originally derived from, though of course their versions are considerably simpler.

-Tony

PR #233, opened 2023-08-21
Added meson based mingw64 compilation and nsis based installer

dov:
The two attached commits do the following:

1. Add cross compilation for a fedora based system for compiling for win64
2. Adds a nsis based windows installer

To compile for windows do:

```
meson setup build_mingw64 --cross-file cross_mingw64.txt -Ddefault_library=shared -Dbuildtype=release
ninja -C build_ming64
ninja -C build_mingw64 installer
```


rfuest:
Nice, thanks for the PR. I didn't have time to test it yet, but I have a few suggestions:
- Please move the windows/nsis specific files in their own directory, like `share/windows`
- I had previously cross compiled GTKWave from Fedora an used their `mingw64-meson` tool to set up the build. Using this should make it unnecessary to ship our own `cross_mingw64.txt` file.
- The README needs updating anyway, but if you have some time it would be great if you could write a section about cross compiling from Fedora to Windows, with all dependencies that need to be installed.

dov:
Hi Ralf. Thanks for the positive feedback! I'll fix the issues and will update the PR.

Regarding the windows/nsis stuff, do you want .ico and the .rc file in that directory as well? I guess yes.

I can't find any fedora mingw64-meson file. Could you please point me to it?

I'll certainly update the README.

I also created an placeholder icon before I discovered that there is an official icon. I'll fix that as well.

src/meson.build
```diff
@@ -128,11 +134,12 @@ gtkwave_include_directories = [
     'cocoa',
 ]
 
-executable(
+exe = executable(
```
rfuest, line 137: The variable name should be more specific because there are multiple executables.
```suggestion
gtkwave_exe = executable(
```

src/pixmaps.c
```diff
@@ -3015,7 +3015,7 @@ GwHierarchyIcons *gw_hierarchy_icons_new(void)
     self->fork = gdk_pixbuf_new_from_xpm_data(arrow_divide);
 
     /* SV */
-    self->interface = gdk_pixbuf_new_from_xpm_data(arrow_inout);
+    self->iface = gdk_pixbuf_new_from_xpm_data(arrow_inout);
```
rfuest, line 3018: Why did you decide to rename this?

src/wave_view.c
```diff
@@ -615,6 +615,9 @@ static gboolean gw_wave_view_draw(GtkWidget *widget, cairo_t *cr)
         if (GLOBALS->disable_antialiasing) {
             cairo_set_antialias(traces_cr, CAIRO_ANTIALIAS_NONE);
         }
+        else
+          cairo_set_antialias(traces_cr, CAIRO_ANTIALIAS_GOOD);
```
rfuest, line 619: Does antialiasing default to none in windows?

gtkwave.nsi
```diff
@@ -0,0 +1,179 @@
...
+  File /r ${SYSROOT}\share\icons\Adwaita
+  
+  SetOutPath $INSTDIR\share\glib-2.0
+  File /r ${SYSROOT}\share\glib-2.0\schemas
+  
+  # Build the gdk-pixbuf.loaders file automatically
+  ExpandEnvStrings $0 %COMSPEC%
+  nsExec::ExecToStack '"$0" /C ""$INSTDIR\bin\gdk-pixbuf-query-loaders" > "$INSTDIR\lib\gdk-pixbuf-2.0\2.10.0\loaders.cache""'
+  
+  ; Set up association with .vcd files
```
rfuest, line 138: What about the other filetypes? At least `.gtkw` and `.fst` should also be associated, in my opinion.

gtkwave.nsi
```diff
@@ -0,0 +1,179 @@
...
+  File ${SYSROOT}\bin\gdbus.exe
+  File ${SYSROOT}\bin\libsigc-2.0-0.dll
+  File ${SYSROOT}\bin\libmpfr-6.dll
+  File ${SYSROOT}\bin\libmpfr-6.dll
+  # Libepoxy in official package is broken, so we use an older working version!
+  # File ${SYSROOT}\bin\libepoxy-0.dll
+  File \usr\local\mingw64\bin\libepoxy-0.dll
+  File \usr\local\mingw64\bin\libfmt.dll
+  File \usr\local\mingw64\bin\libgmp-10.dll
+  File ${OUTDIR}\src\${NAME}.exe
```
rfuest, line 114: The command line only tools from the `src/helpers` directory should also be installed.

rfuest (changes requested):

gtkwave.nsi
dov: I removed this, since it has nothing to do with the windows complation.

However there is still a issue with the paiinting. Part of the problem is that you are drawing through the pattern (where M=MoveTo L=LineTo S=Stroke) MLSMLSMLS... instead of MLLLLLL...S which is the prefered way of using cairo.

rfuest: > However there is still a issue with the paiinting.

What kind of issue, performance or visual artefacts?

> Part of the problem is that you are drawing through the pattern (where M=MoveTo L=LineTo S=Stroke) MLSMLSMLS... instead of MLLLLLL...S which is the prefered way of using cairo.

The current drawing code is very old and hasn't changed a lot since the GTK 1 days. I've done some extensive refactoring of the drawing code a while back, which I originally intended to merge in as part of the GTK 4 port, but I'm considering to merge it in earlier for GTKWave 3.4. The new code should make it easier to optimize the calls that are made to the cairo API.

I had done some tests with batching of strokes as part of #224, but didn't find any significant performance improvement when I try to batch a lot of strokes. Around ~10 lines per stroke seemed to give the best performance. There was some performance anomaly if I tried to draw thousands of lines per stroke, but I didn't investigate that further.

The `MLLLLL...S` pattern is often not possible, because the line color changes between segments. `MLML...S` is the best we could do in these situations.

dov: I was talking about the rendering aesthetics, and not about speed. E.g. the stereo.vcd file (which I found in some discussion list) which I found on some discussion list. I investigated the rendering problems today and found that there are to reasons for the bad rendering.

1. The main reason is that the coordinates are turned into integers before they are passed into `XXX_gdk_draw_line`. This causes lots of x-coordinate  collisions (because multiple time points are rounded to the same x-coordinates). This is rounding interfers with cairo's ability to calculate the pixel coverages.
2. The fact that you stroke after each ML. I agree that doing MLMLMLML...S would be ok. But doing MLSMLS.... causes multiple renderings of the same pixels, and the line become thicker than intended.

I working on an experimental branch where I changed the integer coordinates to double and solved these issues quite hackishly. I hope to upload a new PR with the changes, but if you are working on a GTK4 branch then it might be redundant.

But this has nothing to do with the Windows and the installers. :-) 

[stereo.zip](https://github.com/gtkwave/gtkwave/files/12409788/stereo.zip)

rfuest: > But this has nothing to do with the Windows and the installers. :-)

Yes, let's continue the discussion in the issue you've opened.

meson.build
```diff
@@ -117,3 +117,40 @@ gnome.post_install(
...
+
+  windows_installer = custom_target(
+    'windows_installer',
+    output: 'install-gtkwave.exe',
+    input: files('./compilation/windows/gtkwave.nsi'),
+    build_always_stale: true,
+    command: nsis_cmd,
+    depends: [gtkwave_exe]
+  )
+  alias_target('windows_installer', windows_installer)
```
rfuest, line 155: This caused problems when I tested it:
```suggestion
```

compilation/windows/gtkwave.nsi
```diff
@@ -0,0 +1,201 @@
...
+!include "DumpLog.nsh"
+!insertmacro GetTime
+
+SetCompress force ; (can be off or force)
+CRCCheck on ; (can be off)
+
+LicenseText "Please read and accept the following license:"
+LicenseData "../../COPYING"
+
+BrandingText "GtkWave Installer"
```
rfuest, line 14:
```suggestion
BrandingText "GTKWave Installer"
```

rfuest (changes requested):
I tried to build the installer and got this error message:
```
/usr/bin/makensis -DBUILDDIR=/home/ralf/Projekte/GTKWave/GIT/gtkwave/build_mingw -DSOURCEDIR=/home/ralf/Projekte/GTKWave/GIT/gtkwave -DICON_DIR=../../compilation/windows -DNAME=gtkwave -DNAME_CAP=GTKWave -DICON_NAME=gtkwave_logo -DARCH=x86_64-w64-mingw32 -DSYSROOT=/usr/x86_64-w64-mingw32/sys-root/mingw -DHOST=w64 -DVERSION=v3.4.0 -DLIBGCCDLL=libgcc_s_seh-1.dll .././compilation/windows/gtkwave.nsi
Error: reading stub "/usr/share/nsis/Stubs/zlib-x86-unicode"
Error initializing CEXEBuild: error setting default stub
ninja: build stopped: subcommand failed.
```

I'm not sure what caused this problem and if I'm just missing a dependency. The packages I've installed are listed in a comment below.

rfuest:
You should also rebase this PR onto the new changes that I made to `meson.build` in `master` yesterday to make sure they didn't break the windows build.

src/pixmaps.c
dov: I couldn't get it to compile without it. There seems to be a `interface` define somewhere. I didn't investigate it.

meson.build
dov: If I don't add line 155, the `ninja -t targets` doesn't show `windows_installer`. How did it cause a problem?

README.md
```diff
@@ -128,6 +128,16 @@ The current environment used is modulesets.  Bug 664894 has an interim fix in
...
 `contrib/bundle_for_osx/gtk_diff_against_modulesets.patch` file.
 
+## Notes for cross compilation under Fedora for Windows
+
+```sh
+mingw64-meson -Dbuildtype=release build_mingw64
+cd build_mingw64
+ninja
+ninja windows_installer
+
```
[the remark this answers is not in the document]
dov: Ok, done.

dov:
@rfuest Ok, I resolved out all your comments. However, I have no idea why your makensis command failed. :-(

I did however notice that I in my nsis file referenced some packages in /usr/local that I have have now removed from the .nsi file. I doubt tat is the reason for your failure though.

dov:
Nice changes. :-) Does makensis work for you now?

rfuest:
> Nice changes. :-) Does makensis work for you now?

Kind of, I guess. I haven't tested the generated installer yet and I'm not sure if I haven't broken anything by moving stuff around. So far I've only tested the executable in Wine and not the installer. I did already find some issues without the installer, but I'll need to retest it in Windows.

To get rid of the error I had mentioned previously I had to also install the `mingw32-zlib` package, but I don't know why the 32 bit version is necessary to create a 64 bit installer.

I also had to comment out a bunch of dll files from the installer script. For some of them I'm pretty sure that they aren't necessary, like `libgfortran`, but others, like `librsvg` might still be necessary.

dov:
Interesting. I guess that nsis itself uses the ming32-zlib to create the executable. Perhaps the installer exe is a 32-bit program even though it installs an 64-bit program.

I don't expect you will have any issues running the installer. The only really expected problems are missing dlls, but these are easily added one by one.

We should probably also add an nsis install page with checkboxes for the file types that we want to be associated with GTKWave instead of the hard coded list of three types as of now.

dov:
@rfuest I just tested your latest changes, and there was one small issue that I had to change. Apparently meson failes to do variable substitution if the variable is proceeded by a backslash. Therefore the construct `...\@application_name@` in the `gtkwave.nsi.in` was never substituted. I solved that by moving the declaration back to an nsis variable. (It is still not clear to me, btw, why you prefered using meson configuration with an in file, instead of using the nsis variables.)

I also tested my int→float change from #236 and I'm happy to report that it works fine on Windows as well. :-)

rfuest:
> (It is still not clear to me, btw, why you prefered using meson configuration with an in file, instead of using the nsis variables.)

I meant to explain the reason why I changed this in my previous comment, but somehow forgot. It was an attempt to remove any relative paths from the installer script to make it easier to maintain. But in the end it didn't quite work out as good as I hoped to and just copying the files into the build directory via `fs.copyfile` turned out to be easier. Feel free to change it back if you like.

> Interesting. I guess that nsis itself uses the ming32-zlib to create the executable. Perhaps the installer exe is a 32-bit program even though it installs an 64-bit program.

That might be the case. It is interesting that I can't even run the `makensis` command at all without installing the 32 bit zlib. Perhaps it is just a missing dependency in the fedora package. At the moment I think we should just add `mingw32-zlib` to the dependencies list.

> We should probably also add an nsis install page with checkboxes for the file types that we want to be associated with GTKWave instead of the hard coded list of three types as of now.

That is an excellent idea!

dov:
@rfuest I just rebased on master and then encountered several compilation errors in the windows cross compilation. I solved these quite by brute force, and would appreciate your help in cleaning it up. Especially the following abomination that I added to the top level build.meson file:

```
# Heuristic setting of flags to make cross compilation work.
#
# Ralf: can you clean this up?
if host_machine.system() == 'windows'
    add_project_arguments(
      ['-DHAVE_LIBINTL_H=0',
       '-DHAVE_STRING_H=1',
       '-D__GNU_LIBRARY__'],
       language: 'c')
endif
```

In another earlier commit I picked up your suggestion, and removed the gtkwave.nsi.in file and replaced its functionality with nsis variables.

rfuest:
I've just removed the bundled implementation of `getopt`, which let me remove the additional defines. Current versions of mingw already seem to include all the necessary functions.

But unfortunately the installer is broken again with the same `zlib` error I had before. I'll need to investigate further how to fix this.

dov:
> Does running `gtkwave --help` work on windows? I haven't had time to reboot into windows and under Wine GTKWave doesn't have any console output.

No, it shows nothing. If it writes to stdout you won't see it when `win_subsystem: 'windows'` is turned on. In my GUI programs I typically either have a Help/Command Line menu that describes the command line options, or popup a text window diallog when running with --help.

rfuest:
I was able to get the installer building again by installing `mingw32-nsis` and commenting out the libxml2 dll in the installer script.

Are the icons in the signal tree also missing in Windows? Under Wine GTK complains that it cannot find the XPM pixbuf loader. The GTKWave in the about dialog is also missing, but that isn't a XPM file.

dov:
I'm having a problem with the some of the icons as well, including the application icon (which I have never been able to get to work reliably on Windows) as well as the about icon. Here's a screenshot showing Windows on top and Linux/X11 on the bottom:
gtkw.

I noticed that GTKWave doesn't call `gtk_window_set_icon()`. I have to check if that helps.

Btw, are the yellow dots in the center of the signal view in my image above something that I added by mistake, or is there a setting that toggles them?

rfuest:
I'll look into fixing the app icon, I think the file might simply not get installed and on my Linux system it picks up an icon from another GTKWave installation.

> Btw, are the yellow dots in the center of the signal view in my image above something that I added by mistake, or is there a setting that toggles them?

They are caused by the `vcd_preserve_glitches` setting. The yellow dots mark transitions to the same value as before.

rfuest:
After #243 has been merged the icon should now be displayed correctly in Linux. But GTKWave must be installed for the icon to work correctly, the icon won't show up if GTKWave is started from the build directory. If you install GTKWave to a temporary prefix in your home directory you'll need to add the `share` directory to the `XDG_DATA_DIRS` environment variable.

The icon will still not be visible in windows, because the application icon doesn't get installed. `io.github.gtkwave.GTKWave.svg` from `share/icons` needs to be installed to `${DATADIR}\icons\highcolor\scalable\apps`.

rfuest:
I just tried to compile GTKWave directly in MSYS2 on Windows and ran into lots of issues :disappointed:. Most of the issues were related to missing DLLs, but the compilation of the code did also initially fail.

But I did at least find part of the issue why the icon isn't showing up in the about dialog. I had a typo in my comment yesterday, it should by `hicolor` instead of `highcolor`. But there is also the issue that the files created by `gtk-update-icon-cache` aren't installed. Even with this issues fixed I had to manually add the `share` directory to the `XDG_DATA_DIRS` variable to get the icon working.

rfuest:
> Another unrelated problem is that there is no application icon (for minimize, and when launching the application). I had a look at the windows `build.ninja` file, and realized that the gtkwave.rc compiled object is not listed in the `gtkwave.exe` link command line. It is compiled, but never linked into the executable. Do you have any idea why?

The `windows` subdir is included into the main `meson.build` after the `src` subdir. Because of this the changes to  `additional_gtkwave_sources` aren't included in the executable. If you move the resource compilation to `src/meson.build` it should work.

PR #243, opened 2023-08-27
Add new application icon

rfuest:
This PR replaces the GTKWave icon and removes the old MIME type icons.


dov:
@rfuest Two issues: I copied the svg icon to $INSTDIR\share\icons\highcolor\scalable\apps but it still does not appear in help about dialog. When I'll have time I'll check if it makes a difference if the icon is in png instead of svg.

Another unrelated problem is that there is no application icon (for minimize, and when launching the application). I had a look at the windows `build.ninja` file, and realized that the gtkwave.rc compiled object is not listed in the `gtkwave.exe` link command line. It is compiled, but never linked into the executable. Do you have any idea why?

PR #244, opened 2023-08-28
Remove old ifdefs and rcvars

rfuest:
This PR removes an old unused ifdef, a global variable, and two rc variables. All of them were previously discussed in different issues and they should all be obsolete and safe to remove.


gtkwave:
BTW, related to obsolete ifdefs...
https://www.muppetlabs.com/~breadbox/software/cppp.html
For future reference, this tool is great, especially with deeply nested ifdefs, misleading indentation, etc.  I've used it in the past to convert GTK2 to GTK3 to remove obsolete GTK-1.3 code.

-Tony

PR #259, opened 2023-09-30
Fix waveview scaling issue when on HiDPI displays

randomplum:
Currently when gtk_widget_get_scale_factor() was !=1 crated surface is not sufficient size, which leads to only 1/scale^2 of the waveview being visible.
I've reproduced it on Linux and MacOS. This patch fixed it on both.

### DIFF
--- a/src/gw-wave-view.c
+++ b/src/gw-wave-view.c
@@ -671,6 +671,7 @@ static gboolean gw_wave_view_draw(GtkWidget *widget, cairo_t *cr)
 static void gw_wave_view_size_allocate(GtkWidget *widget, GtkAllocation *allocation)
 {
     GwWaveView *self = GW_WAVE_VIEW(widget);
+    int scale;
 
     GTK_WIDGET_CLASS(gw_wave_view_parent_class)->size_allocate(widget, allocation);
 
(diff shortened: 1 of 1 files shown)


rfuest (approved):
LGTM, thanks for the PR.

rfuest:
> I've reproduced it on Linux and MacOS. This patch fixed it on both.

How did you reproduce the issue on Linux? I wasn't really able to reproduce it when I tried a few days ago. Are you running X11 or Wayland and which DE/WM are you using?

randomplum:
> > I've reproduced it on Linux and MacOS. This patch fixed it on both.
> 
> How did you reproduce the issue on Linux? I wasn't really able to reproduce it when I tried a few days ago. Are you running X11 or Wayland and which DE/WM are you using?

Latest fedora with Wayland/Gnome on 4k screen, seting scaling to 200% made gtk_widget_get_scale_factor return 2 and I got the same 1/4 view

rfuest:
OK, thanks for the info. I'm also using GNOME and Wayland and was now able to reproduce the issue and can also confirm that the fix works. No idea what I did wrong the last time I tried to reproduce the issue.

Issue #250, opened 2023-09-15
Unable to launch GTKWave on macOS Sonoma


andylin2004:
<img width="328" alt="Screenshot 2023-09-15 at 8 55 51 AM" src="https://github.com/gtkwave/gtkwave/assets/60021592/18aa7fc9-f5a2-4485-958e-29d82691393d">

This prompt will show up, and then the app will close out

gtkwave:
Simply showing a screenshot doesn't convey much information.  What is the version of gtkwave and where did you get it?
Is this with respect to the binaries grabbed off the sourceforge site?  If so, sorry, I don't own a Mac.  Those binaries were compiled on a machine running a Snow Leopard (10.6) VM created from back around 2011 or so.  I stopped creating binaries when it looked like compatibility issues would start occurring.

Thanks,
-Tony

andylin2004:
I am using version 3.3.107, installed via brew on macOS.

gtkwave:
OK, you're using the most recent version that I built...which was three years ago.
As I said, I don't have a Mac, so unfortunately, I can't build a binary that will work for you.  Setting up a build environment on Mac using GTK-OSX is a fairly involved process unless it's been simplified.  Another user might be able to help you.

-Tony

randomplum:
@andylin2004 

if you're using brew you can try:


class Gtkwave < Formula
  desc "GTKWave"
  homepage "https://gtkwave.sourceforge.net"
  license "GPL-2.0-or-later"
  head "https://github.com/randomplum/gtkwave.git", branch: "macos_sonoma"

  depends_on "desktop-file-utils" => :build # for update-desktop-database
  depends_on "shared-mime-info" => :build 
  depends_on "gobject-introspection" => :build
  depends_on "meson" => :build
  depends_on "ninja" => :build
  depends_on "pkg-config" => :build
  depends_on "gtk+3"

  def install
    ENV["DESTDIR"] = "/"
    system "meson", "setup", "build", *std_meson_args
    system "meson", "compile", "-C", "build", "--verbose"
    system "meson", "install", "-C", "build"
  end

end

save it in gtkwave.rb file and run brew install --build-from-source --HEAD ./gtkwave.rb

This is the only needed part of the patch I have in the fork
<img width="754" alt="regex" src="https://github.com/gtkwave/gtkwave/assets/58212048/d8a2927e-9417-4dd1-8c9b-9f110a157dff">


[...]

PR #260, opened 2023-09-30
Update macos support

randomplum:
This change updates macos support.


rfuest:
How is the menu bar handled in the macOS build? In the screenshots in #250 there seems to be a menu bar inside the window, but there is also some code to use the global macOS menubar. I don't think there should be both and only using the global menu bar should be more familiar to macOS users.

Regarding the `getopt` files: I think we can remove the local copies of getopt entirely and rely on the system files instead. Every platform we compile for should have the necessary functions and at some point we will move from getopt to using the glib command line handling anyway.

randomplum:
> How is the menu bar handled in the macOS build? In the screenshots in #250 there seems to be a menu bar inside the window, but there is also some code to use the global macOS menubar. I don't think there should be both and only using the global menu bar should be more familiar to macOS users.
> 
> Regarding the `getopt` files: I think we can remove the local copies of getopt entirely and rely on the system files instead. Every platform we compile for should have the necessary functions and at some point we will move from getopt to using the glib command line handling anyway.


With full mac integration enabled it looks like that (different icons are due to a different machine with gtk theme set to not default) :

image

Previous screenshot was from early experiments without any specific macos code enabled, then the menu bar was only in the application (gtk on macos nicely handles both ways)


ok, I'll remove the getopt files. I disabled them for mac since clang 15 with default flags really doesn't like the old-school function parameter types definitions.

rfuest:
`gnu-getopt.h` can also be removed and replaced by `#include <getopt.h>`.

randomplum:
> `gnu-getopt.h` can also be removed and replaced by `#include <getopt.h>`.

done

rfuest (approved):
Looks good.

PR #261, opened 2023-10-01
Fix "grab to file" on HiDPI displays

randomplum:
gdk_pixbuf_get_from_window is scale aware, but cairo_image_surface_create is not.

There are some calls to cairo_image_surface_create in mousover, but I'm not familiar with the project yet to know if those are also affected.

### DIFF
--- a/src/menu.c
+++ b/src/menu.c
@@ -3336,13 +3336,15 @@ void menu_write_screengrab_cleanup(GtkWidget *widget, gpointer data)
     gw = gtk_widget_get_window(GTK_WIDGET(GLOBALS->mainwindow));
     if (gw) {
         GtkAllocation allocation;
+        int scale = gtk_widget_get_scale_factor(GTK_WIDGET(GLOBALS->mainwindow));
         gtk_widget_get_allocation(GLOBALS->mainwindow, &allocation);
 
         pb = gdk_pixbuf_get_from_window(gw, 0, 0, allocation.width, allocation.height);
 
         if (pb) {
             cairo_surface_t *surface =
-                cairo_image_surface_create(CAIRO_FORMAT_RGB24, allocation.width, allocation.height);
+                cairo_image_surface_create(CAIRO_FORMAT_RGB24, allocation.width * scale,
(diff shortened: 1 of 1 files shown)


rfuest (approved):
LGTM

rfuest:
> There are some calls to cairo_image_surface_create in mousover, but I'm not familiar with the project yet to know if those are also affected.

The mouseovers shouldn't use custom drawing at all, the code for the mouseovers is much more complicated than it needs to be. I'll create a PR to change them to use a simple `GtkLabel` instead.

gtkwave:
"The mouseovers shouldn't use custom drawing at all, the code for the mouseovers is much more complicated than it needs to be. I'll create a PR to change them to use a simple GtkLabel instead."

If I remember correctly, the reason surfaces (or windowless pixmaps) were used, it was because it was impossible in GTK2 to have a GtkLabel track the mouse pointer's movement dynamically, or have it track the movement dynamically in a sane way.  I don't know what the case is with GTK4.

-Tony

rfuest:
> If I remember correctly, the reason surfaces (or windowless pixmaps) were used, it was because it was impossible in GTK2 to have a GtkLabel track the mouse pointer's movement dynamically, or have it track the movement dynamically in a sane way.  I don't know what the case is with GTK4.

I'm not sure about that, it looks like the tracking of the mouse movements has always been handled by moving the mouseover's `GtkWindow` around. It shouldn't matter if the content of the `GtkWindow` was drawn manually or by some other widgets. But in the end the exact reasons why it was implemented this way doesn't matter. At least in GTK 3 it is possible to use `GtkLabel`s instead, see PR #263.

gtkwave:
Attempting to remember more, I want to say it was something quite odd such as not being able to extract absolute x/y coordinates for the window manager, so something relative to the main window was required (hence gdk_window_get_origin()/gtk_window_move() usage in move_mouseover()), and labels or whatever else I used failed to be located properly on the window manager I was using at the time (probably Fvwm95).  OK, good, if it works it works.  For all I know, my window manager was being weird.

-Tony

PR #264, opened 2023-10-02
Move additional types to libgtkwave

rfuest:
This PR moves a bunch of types to libgtkwave to make them available to be use inside libgtkwave.

### DIFF
--- a/lib/libgtkwave/src/gtkwave.h
+++ b/lib/libgtkwave/src/gtkwave.h
@@ -9,3 +9,4 @@
 #include "gw-color.h"
 #include "gw-color-theme.h"
 #include "gw-hist-ent.h"
+#include "gw-vector-ent.h"
--- a/lib/libgtkwave/src/gtkwave.h
+++ b/lib/libgtkwave/src/gtkwave.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "gw-types.h"
 #include "gw-time.h"
 #include "gw-named-markers.h"
 #include "gw-marker.h"
@@ -10,3 +11,5 @@
 #include "gw-color-theme.h"
 #include "gw-hist-ent.h"
 #include "gw-vector-ent.h"
+#include "gw-node.h"
+#include "gw-fac.h"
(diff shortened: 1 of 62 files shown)


sequencer:
any plan to move more things to libgtkwave? e.g. fst/vcd loader.

rfuest:
> any plan to move more things to libgtkwave? e.g. fst/vcd loader.

Yes, that's the goal. In my private branch the loaders are already moved to libgtkwave, but I'll need some time to move everything over to libgtkwave without breaking GTKWave.

rfuest:
> Is possible to create a gw_project from my own project...

All non UI components of GTKWave will eventually end up in libgtkwave. It will be possible to create a new `GwProject` from a dump file (e.g. FST/VCD) or by loading a GTKWave savefile (`.gtkw` or the new JSON format).

> ...to redesign the UX after this migration is done?

What are you trying to achieve that you think cannot be integrated into GTKWave itself? IMO it would be a better use of developer resources to focus on improving GTKWave instead. Also keep in mind that libgtkwave is an internal library at the moment and there won't be any guarantees that new versions won't break API/ABI compatibility initially.

sequencer:
> What are you trying to achieve that you think cannot be integrated into GTKWave itself? IMO it would be a better use of developer resources to focus on improving GTKWave instead. Also keep in mind that libgtkwave is an internal library at the moment and there won't be any guarantees that new versions won't break API/ABI compatibility initially.

Actually the current issue for us in the tcl RPC...

I wanna create an IDE plugin to read from our DSL code(Chisel), and IDE(VSCode/Intellij) send RPC to a debug server, then debug server read xml from verilator and finally send RPC to gtkwave. So this will be a totally UX which will tightly integrated with an IDE, and using its own hierarchy viewer and schematic viewer.

rfuest:
Sounds interesting.

> Actually the current issue for us is the tcl RPC, wish we can have a json RPC instead...

The current plan is to use D-Bus for the RPC in future versions of GTKWave. If D-Bus doesn't work for your application, and JSON RPC is the only convenient option, it should be possible to add a custom RPC server for your application as a GTKWave plugin, when the plugin interface is finished.
If you have any suggestions for improvements to the RPC interface please open a new issue.

Issue #280, opened 2023-11-01
Enormous RAM usage with big wave files.


albydnc:
I have a wavefile in GHW format that is ~300 MB in size.
I have tried to open it with gtkwave and the tool hangs requesting tons of memory ( I have a 16 GB machine and it takes 20 GB requiring the os to swap).
It never opens the GUI.

rfuest:
All traces in a GHW file are loaded at once and converted into a format that is optimized for displaying in GTKWave, which causes the large memory consumption. If you can use FST files instead of GHW files you should try to use that format instead. Traces in FST files are lazily loaded when they are needed.

albydnc:
Unfortunately, the simulator I'm using is not able to export some signals into the FST format (see [issue on GHDL](https://github.com/ghdl/ghdl/issues/2527#issue-1973128643) ), so I need to use GHW.

rfuest:
I'm afraid there might not be much we can do to fix this in the short term. In the long term it will be possible to improve GTKWave's memory consumption for GHW files, but these changes will need to wait until the current refactoring is finished.

Can you provide the GHW file, or one with similar characteristics, for further testing? My test file did "only" consume about ~15 times the GHW filesize as the peak memory consumption. Your file is much worse than that and it would be interesting to check if the high memory consumption is just caused by the expected inefficiencies or if something else is wrong.

albydnc:
Thanks for the help! Here is the GHW file in question. https://cernbox.cern.ch/s/GBA8otWWO0MgeqR

rfuest:
Thanks for the file, getting access to actual dump files helps a lot to improve GTKWave.

The GHW file contains 35.649.163 signals, which causes GTKWave to use ~14 GB of memory just to store the hierarchy. The GHW loader seems to create one hierarchy entry for each bit in the design, which consumes a lot of memory. A huge chunk of this memory is consumed by storing the full name of each of the 35 million signals. Fixing this will require some major changes to the GHW loader and/or the way GTKWave stores the hierarchy.

[...]

PR #300, opened 2024-01-18
libghw - Update to latest upstream version, speedup

stefanlippuner:
This updates libghw to the upstream from https://github.com/ghdl/ghdl. The main change is skipping a linear search in `ghw_read_cycle_cont`, if this is possible. In my testing this provided a massive speedup. With this change, loading the ghw file from https://github.com/gtkwave/gtkwave/issues/280#issuecomment-1791391127 took about 20s instead of more than 6 minutes before. The file consumes about 29 GB of RAM, and behaves quite strangely (e.g., signals change name after inserting them to the wave window). So there may still be some bugs left, but the results from the two implementations of `ghw_read_cycle_cont` matched (tested with an assert for a few different log files).

Includes the following changes:
 - https://github.com/ghdl/ghdl/pull/1825
 - https://github.com/ghdl/ghdl/pull/1807
 - https://github.com/ghdl/ghdl/pull/2585

Related to #280

### DIFF
--- a/lib/libghw/libghw.c
+++ b/lib/libghw/libghw.c
@@ -655,7 +655,7 @@ ghw_read_type (struct ghw_handler *h)
       if (t == EOF)
 	return -1;
       if (h->flag_verbose > 1)
-	printf ("type[%d]= %d\n", i, t);
+	printf ("type[%u]= %d\n", i, t);
       switch (t)
 	{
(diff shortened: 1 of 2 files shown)


tbybell:
If Tristan has blessed this upstream, I'm good with it.

PR #302, opened 2024-01-20
Dump file loaders refactoring

rfuest:
This PR moves the dump file loaders into libgtkwave. I've marked it as a draft, because it still needs some work and only the GHW loader is currently moved into libgtkwave. The other loaders still use the GLOBALS struct and need further refactoring to make them work standalone.


rfuest:
While this isn't 100% finished, I've decided to merge this now and finish it in smaller PRs. This will make it easier for others to work on the codebase at the same time.

FurryAcetylCoA:
When building in a clean environment, it throws the following error:
```
In file included from ../lib/libgtkwave/test/dump.c:1:
../lib/libgtkwave/src/gtkwave.h:4:10: fatal error: gw-enums.h: No such file or directory
    4 | #include "gw-enums.h"
      |          ^~~~~~~~~~~~
compilation terminated.
```
Building again with the same command makes this problem disappear.
It seems there is a build order issue. `test/dump.c` gets built before `gw-enums.h` is generated.

Thanks  again for your significant contribution for this project.

rfuest:
Thanks for reporting this, I had only encountered this issue once before when CI failed in one of my recent PRs. But it did work afterwards and I didn't investigate further at the time. I hope that #351 adds the necessary build time dependencies to get the order right.

Issue #209, opened 2023-06-15
RFC: New savefile format


rfuest:
The current savefile format isn't easy to read or to generate by an external tool. I would like to introduce a new file format for GTKWave 4. At the moment I'm planning to use JSON to store the save file. While I'm not a huge fan of JSON myself it has the advantage that it is very common and can easily be used in many programming languages or modified by tools like `jq`. And with a modern editor with JSON support it isn't too bad to edit by hand.

In a previous discussion about a new savefile format TCL has been suggested as a new file format, but I prefer a serialization format like JSON for two reasons. The first are possible security implications of using a scripting language as a savefile format. While it is unlikely that this would be exploited, I just don't like the idea of allowing arbitrary code execution when opening a file. The second reason is that it wouldn't be possible to preserve custom code in the TCL file if a savefile is overwritten by GTKWave. If, for example, a user has added signals in a loop inside the TCL file and then overwrites the file using "Save" in the UI any custom code would be gone. Using TCL as a savefile has it's use cases, like https://github.com/gtkwave/gtkwave/pull/114, but I think it shouldn't be the primary savefile format and instead be a secondary format that can be opened read only and exported.

## JSON Example

This example shows a rough idea of what I have in mind for a JSON format.

![grafik](https://github.com/gtkwave/gtkwave/assets/226123/16fe4f69-acf2-49d4-93ec-b6766d92751b)

### Current format

```
[*]
[*] GTKWave Analyzer v3.3.115 (w)1999-2023 BSI
[*] Thu Jun 15 17:32:40 2023
[*]
[dumpfile] "/home/ralf/git/gtkwave/gtkwave/gtkwave3-gtk3/examples/des.fst"
[dumpfile_mtime] "Mon Oct 11 15:44:12 2021"
[dumpfile_size] 158102
[savefile] "/tmp/old.gtkw"
[timestart] 0
[size] 1000 600
[pos] 1881 387
*-3.864544 5 8 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1
[markername] AMarker
[treeopen] top.
[treeopen] top.des.
[sst_width] 273
[signals_width] 94
[sst_expanded] 1
[sst_vpaned_height] 157
@29
top.clk
@22
top.ct[1:64]
@10420
top.i
@20000
-
@800200
-round1
@22
[color] 1
[...]

PR #306, opened 2024-02-07
Ported colors from the Vivado Simulator

radomirvr:
This introduces 20 new colors from the Xilinx Vivado Simulator color palette. The order of colors in the menu is also brought over, but the color indices (of the colors that already existed) remain the same. This also means that the "Menu" order and the "Color Cycle" order no longer match.


rfuest (changes requested):
Thanks for the PR. I like the idea of having more colors available, but I think we need to make a few tweaks before this can be merged.

When you use the color cycle feature a lot of similar gray colors are adjacent to each other:
grafik
The purpose of the color cycle feature is to quickly make adjacent traces easier to distinguish, which isn't the case if they are similar shades of gray. I think we need to use a custom order, or perhaps even only a subset of the available colors, to make adjacent traces easier to distinguish.

The new colors are also not available in the trace properties dialog: https://github.com/gtkwave/gtkwave/blob/51aa052db9440d634f588178759f20fec4cffdc2/src/showchange.c#L48-L56
With the larger number of colors we should refactor the code a bit, so that we don't have to maintain lists of colors all over the codebase. I would suggest to change the `RAINDOW_COLORS` array to also include the name of the color and to add a `gw_rainbow_color_get_name` function.

radomirvr:
I agree that a refactoring is needed regarding the colors, as the colors are defined in multiple places. Honestly, I like the way that Xilinx handled that - the `.wcfg` file (equivalent to `.gtkw`) stores the #RRGGBB value of the selected color for every signal.

I also think that selecting only a subset of the colors for the color cycle could solve that issue, and that seems to be an easy fix (reordering the color indices and changing the value of `WAVE_NUM_RAINBOW`?).

tbybell:
Looks interesting.  I agree with Ralf's comments.  I myself use color sparingly, mostly so I can quickly spot the valid signal for a group of other related signals, but whatever is good for other users sounds good.

I'd say don't be afraid to break backward save file compatibility in this area.  The old color selection and ordering is completely arbitrary.  Indigo, for example, seems useless with respect to contrasting color, but was added only because it's listed in a traditional ROYGBIV palette.

rfuest:
> I agree that a refactoring is needed regarding the colors, as the colors are defined in multiple places. Honestly, I like the way that Xilinx handled that - the `.wcfg` file (equivalent to `.gtkw`) stores the #RRGGBB value of the selected color for every signal.

Storing the colors differently is something I would like to implement in the new savefile format (#209). But there are advantages to use named colors (or indices) instead of saving the RGB values in the savefile. If we decide to implement support for color themes, each theme could provide a custom set of trace colors to make sure that all colors have a good contrast to the background color. This would be especially important for light color themes. But we could still provide a way to also allow custom RGB colors in the savefile.

> I also think that selecting only a subset of the colors for the color cycle could solve that issue, and that seems to be an easy fix (reordering the color indices and changing the value of `WAVE_NUM_RAINBOW`?).

I would prefer to decouple color cycling and the color indices completely. Here is a method that could be added to `gw-color.c` to implement color cycling:
```c
GwRainbowColor gw_rainbow_color_cycle(GwRainbowColor self)
{
    static const GwRainbowColor COLOR_CYCLE[] = {
        GW_RAINBOW_COLOR_RED,     
        GW_RAINBOW_COLOR_GREEN, 
    };

    for (gsize i = 0; i < G_N_ELEMENTS(COLOR_CYCLE); i++) {
        if (COLOR_CYCLE[i] == self) {
            return COLOR_CYCLE[(i + 1) % G_N_ELEMENTS(COLOR_CYCLE)];
        }
    }

    return COLOR_CYCLE[0];
}
``` 
The `COLOR_CYCLE` array will need to be adjusted and is just an example. To use the method the `color_prev++` line in `colorformat` in `menu.c` needs to be changed into: `color_prev = gw_rainbow_color_cycle(color_prev - 1) + 1;`

radomirvr:
I'd honestly like to dive deeper here, attempt a proper colors cleanup including savefile modification.

Also, I imagine that storing the RGB value instead of the color index in the trace could be beneficial.
That would also allow us to create a menu where the user would be able to specify the colors along with their respective order, while maintaining compatibility with the `.gtkw` files created by others. That may be overkill for a feature such as trace colors, but seems easy enough to implement.

What do you think?

rfuest:
I like the idea, but I'm not sure that now is the right moment to implement this. My plan for GTKWave 4 is to only support the current savefile format for loading and always save in the new JSON based format. If you implement the changes to the savefile format now they would only be available to users of the GIT version, which might be a waste of time. But any help with designing a flexible color in the new format would be highly appreciated.

radomirvr:
Fair enough, I was not aware of the version 4 roadmap. Still, I consider the RGB superior to color index.

rfuest:
> Still, I consider the RGB superior to color index.

The JSON savefiles will use names for the predefined colors. This is easier to use than RGB values, if you generate savefiles with an external tool, or if you manually edit a savefile. `"color": "spring-green"` is, in my opinion, much easier to understand than  the RGB value `00FF80`.

But, as I mentioned before, there is also room to support custom RGB values for traces. In addition to the named colors the savefile could also support other types of color value, like `"color": "#FF0000"` or `"color": "rgb(255, 0, 0)"`.

tbybell:
Historically, color names were used in gtkwave because in the mid 90s or so, the Sun workstations I was using at the university were 256 color and most of the X11 command like tools for changing things like background color used well-defined names because RGB values are hard to read in a config file.  See https://en.wikipedia.org/wiki/X11_color_names which should be a good representation of the origin.

At least nobody here is going to be arguing for HSV values...  :-)

radomirvr:
> This is easier to use than RGB values, if you generate savefiles with an external tool, or if you manually edit a savefile. `"color": "spring-green"` is, in my opinion, much easier to understand than the RGB value `00FF80`.

I don't know, anyone doing such low-level manipulations with a savefile should already be familiar with RGB values. This may be a biased opinion, but it seems to me that would also avoid mapping color strings to enums/ints/RGB.

> Historically, color names were used in gtkwave because in the mid 90s or so, the Sun workstations I was using at the university were 256 color and most of the X11 command line tools for changing things like background color used well-defined names because RGB values are hard to read in a config file. See https://en.wikipedia.org/wiki/X11_color_names which should be a good representation of the origin.

Oh, that makes a lot more sense now. I guess the values were copied over instead of calling X11 libraries for the sake of Windows compatibility?

> At least nobody here is going to be arguing for HSV values... :-)

Yeah, no way I would suggest such a horror :). My biggest complaints are about using floating point values for RGBA, of which alpha is never used; and having color defines in multiple places...

rfuest:
> I don't know, anyone doing such low-level manipulations with a savefile should already be familiar with RGB values. This may be a biased opinion, but it seems to me that would also avoid mapping color strings to enums/ints/RGB.

I don't doubt that they would be familiar, but that is not the point. Picking from a limited set of names is much easier than having to come up with a RGB value. If we create a JSON schema the list of available colors would even be shown during autocompletion in a JSON editor.

Here is how I think we should implement colors in GTKWave 4. This allows the usage of custom colors without making it harder for users who don't need to pick from hundreds or millions of colors and just want to quickly make a trace distinct:
- We have about 10 predefined colors, that work similarly to the current colors.
  These will be referenced be name in the savefile. These names are not the same as X11 or CSS color names, because the predefined colors are themeable. What I mean by that is that the predefined color with the name "green" won't necessarily always map to `00FF00`, but is defined in a color theme. This ensures that the predefined colors will always be displayed with a good contrast to the background, which is e.g. important when you want to export a screenshot with a light background.
- All other colors are custom colors.
  In the save file format these will be stored as an RGB hex value `#RRGGBB`. Custom colors will not be affected by a changed color theme and the user has to ensure that the contrast to the background is maintained. While it wouldn't be hard to also support X11 or CSS color names for custom colors I think that it isn't necessary
- A "Custom..." color option will be added underneath the predefined colors in the trace color selection.
  This will pop up the GTK color picker dialog, which has a nice selection of predefined colors and also allows the user to store their own custom colors. This should be pretty comfortable to use and we don't need to write additional code in GTKWave to store a custom set of colors. But I'm not sure if this works the same on all platforms or if an OS specific color is used in Windows and macOS.

tbybell:
Sounds good.  At this point, I'd think the RGB selection widget as used by GTK on all platforms is the same.  I think that's one of the standard widgets it compiles for its examples once the toolkit has been built.

PR #318, opened 2024-03-19
Correctly handle the return value of shmat

FurryAcetylCoA:
I was working on `rtlbrowse` and noticed the return value of shmat was being incorrectly handled
The current implementation cause seg fault with command  `rtlbrowse 0a123`  (or any hexadecimal digits that do not represent a valid shared memory ID)

According to POSIX 2008 and [man page](https://linux.die.net/man/2/shmat) of shmat
shmat will return (void *) -1 upon failure, not 0

I also corrected this issue for  other shmat in the code base.  
Although `MapViewOfFile` does return 0 upon failure. It already got its own error handler wrapped in `ifdef __MINGW32__`

### DIFF
--- a/contrib/rtlbrowse/stem_recurse.c
+++ b/contrib/rtlbrowse/stem_recurse.c
@@ -306,7 +306,7 @@ if(i==len)
 #else
 	anno_ctx = shmat(shmid, NULL, 0);
 #endif
-	if(anno_ctx)
+	if(anno_ctx != (void *) -1)
 		{
 		if((!memcmp(anno_ctx->matchword, WAVE_MATCHWORD, 4))&&(anno_ctx->aet_type > WAVE_ANNO_NONE)&&(anno_ctx->aet_type < WAVE_ANNO_MAX))
 			{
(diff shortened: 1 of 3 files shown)


rfuest (approved):
Thanks, this looks good. `(void *)-1` is an unusual choice for an error value.

In the long term I would like to replace this custom IPC mechanism with D-Bus to make it more robust.

rfuest (changes requested):
Sorry, I just noticed an issue: `anno_ctx` and `dual_ctx` must now be initialized to `(void*)-1` in globals.c.

rfuest:
No, wait that is wrong again, I really shouldn't review code today. The globals must be `NULL` if `anno_ctx` or `dual_ctx` aren't used. The result of the `shmat` call needs to be assigned to a temporary variable fist and the globals should only be updated if the result is not `(void*)-1`.

FurryAcetylCoA:
Thanks for your reply.
IMHO, considering that the program will continue to execute when `anno_ctx` is not used. For `anno_ctx` in `main.c` a temporary variable is indeed necessary. 
However, regarding `dual_ctx`, since the program will terminate with `exit(255)` upon failing to attach the shared memory segment. I believe that introducing a temporary variable for it isn't necessary.

rfuest (approved):
LGTM

PR #332, opened 2024-04-24
Fix memory leak in filter loads

FurryAcetylCoA:
Previously, if the executable file was not found, the `cmd` pointer would not be properly released, leading to a memory leak.

`// I also recommend adding a popup notification when the executable file is not found. This would reduce confusion for users. However I'm not sure how to notifications get popped in gtkwave. 
`

### DIFF
--- a/src/ptranslate.c
+++ b/src/ptranslate.c
@@ -152,14 +152,13 @@ static void load_proc_filter(int which, char *name)
 
     result = fscanf(stream, "%s", abs_path);
 
-    if ((strlen(abs_path) == 0) || (!result)) {
-        status_text("Could not find filter process!\n");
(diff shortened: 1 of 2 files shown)


rfuest (approved):
LGTM, thanks for the PR.

rfuest:
> // I also recommend adding a popup notification when the executable file is not found. This would reduce confusion for users. However I'm not sure how to notifications get popped in gtkwave.

This is a good idea. Giving better feedback about problems that occurred when a project is loaded is planned for GTKWave 4. I'll open a new issue to track this improvement.

tbybell:
Thx, I'll fold this into LTS in a minute.

PR #340, opened 2024-04-30
[docs] move one hierarchy level to the left sidebar, by splitting second level sections into separated pages

umarcor:
Ref: #144

In order to reduce scrolling and to have a better overview of the content through the left sidebar (since the toctrees are hidden and a complete index would be too long anyway), this PR splits second level sections into separated pages.
Toctrees' `caption` feature is used to show "chapters" in the sidebar.

Other than that, the content is unmodified.


rfuest (approved):
That already looks much better. I think we should tweak it some more in the future, but I'll merge this PR now.

PR #349, opened 2024-05-28
Add GitHub Action workflow 'docs' using new recommended way 

FurryAcetylCoA:
I noticed that the PR #338 hasn't been updated in a while.
This PR implements the same GitHub action as #338 but uses the new recommended way of publishing to GitHub Pages.


rfuest (approved):

rfuest:
Great, we now have a new documentation page: https://gtkwave.github.io/gtkwave/

@tbybell: It might be a good idea to add a link from the SF page to the new docs, or we could even start to migrate to the new page entirely.

PR #350, opened 2024-05-31
Remove stray code related to AET2 and other deprecated formats

FurryAcetylCoA:
Since the only define of `WAVE_ARRAY_SUPPORT` is conditioned on `AET2_IS_PRESENT`, and AET2 support is removed. 
It should be safe to remove the associated code.

Additionally, remove the references to "LXT2/VZT" in print_help


rfuest:
I've just decided to merged #302. While the refactoring of the loaders isn't quite finished yet I think it's better to base any further work on that code. I can review your changes after you've rebased the PR.

tbybell:
It's totally safe to remove that code as a lot of it is stuff like adding "{index}" signal importing.  AET2 could store arrays natively in the dumpfile instead of blasting them out into individual rows of signals, but it was the only file format that allowed that.  In the future, if such a thing would ever exist again (e.g., some compelling new format can support that), it would be easier to have an external array viewer window indexed by marker time or whatever rather than put it up in main window.

AET2_ALIASDB_IS_PRESENT is the only other real AET2 feature that was interesting, but that's not needed either as signals in modern formats have their aliases/copies in the file itself, so an external file is not needed.

FurryAcetylCoA:
Blocked by https://github.com/gtkwave/gtkwave/pull/302#issuecomment-2143304750

src/analyzer.c
```diff
@@ -535,23 +535,7 @@ int AddNode(GwNode *nd, char *aliasname)
...
-        unsigned int i;
-        int rc = 1;
-
-        for (i = 0; i < nd->array_height; i++) {
-            rc |= AddNodeTraceReturn(nd + i, aliasname, NULL);
-        }
-        return (rc);
-    }
-#endif
+    return (AddNodeTraceReturn(nd, aliasname, NULL));
```
rfuest, line 538: `AddNodeUnroll` was only used for the array support. We could simplify the code a bit by removing it entirely and by replacing 'AddNodeUnroll` with `AddNodeTraceReturn` everywhere it is used.

src/savefile.c
```diff
@@ -82,19 +82,7 @@ char *extract_dumpname_from_save_file(char *lcname, gboolean *modified, int *opt
...
-    {
-        strcpy(GLOBALS->buf_menu_c_1, hname);
-    }
-#ifdef WAVE_ARRAY_SUPPORT
-    else {
-        sprintf(GLOBALS->buf_menu_c_1, "%s{%d}", hname, n->this_row);
-    }
-#endif
-
+    strcpy(GLOBALS->buf_menu_c_1, hname);
```
rfuest, line 85: The same as `AddNodeUnroll` also applies here. But the required changes might be a bit more complex than just replacing `append_array_row(n)` with `n->nname` everywhere, because we need to ensure that the code that uses the returned string doesn't modify it. I have been caught out by issues where a string has been unexpectedly modified in the past and the compiler didn't catch the issue because the GTKWave codebase doesn't use the `const` keyword to prevent this.

rfuest (approved):
LGTM. I've left two suggestions that could improve the code further, but we can also merge this PR without these changes if you prefer.

src/analyzer.c
FurryAcetylCoA: Thanks for your advice.

I just noticed that `AddNode` and `AddNodeUnroll` are now identical.
For now, I prefer replacing `AddNodeUnroll` with `AddNode` since its name is more straightforward and consistent with other `Add` functions in `analyzer.c` (`AddNode`, `AddVector`).

However, I'm still trying to figure out the relationship between `node` and `trace`, and what they stand for. These two data structure come from the CVS era, so I can't find much information why they are added.

Also, does the "return" in `AddNodeTraceReturn` mean "Add a node, **then** return the trace (via `GwTrace **tret`)"?

tbybell: Node is what the loaders will generate.
Trace wraps more data around that to make the node information useful to be viewed onscreen.
IIRC, BitVector is generally used also for bit extraction/concatenation from a Node.  In a Trace, there is a union that will contain either a Node pointer or a BitVector pointer.

rfuest: I also had some issues understanding how this part of the code worked at first. `AddTrace` does heavily rely on the global state in `GLOBALS`, which makes code like the savefile loader hard to understand.

I haven't spend a lot of time to clean up that part of the code, because I'm planning to replace the current `GwTrace`  (aka `TraceEnt`, `Trptr` in LTS) with a new GObject based type. I've already done some work on this and implemented a new parser for the `gtkw` savefile format, but that hasn't been merged yet.

src/savefile.c
FurryAcetylCoA: Great, I am working on this.

rfuest:
I'll merge this PR now, because the changes in this PR will make it easier for me to move some code to libgtkwave. Thanks for that!

Please open a new PR If you still want to work on removing `append_array_row`.

PR #353, opened 2024-06-02
Fix leading [LF] skipping issue

FurryAcetylCoA:
When reading DOS format files. "AB[_CR_][_LF_]CD" should break to "AB" and "CD"  not  "AB" and "[_LF_]CD". 
The later one will mess up many function including `read_save_helper` .
`continue` must be added or leading [_LF_](or [_CR_]) will be incorrectly appended to the start of `line`


rfuest (approved):
LGTM. Good catch!

PR #354, opened 2024-06-02
Fix Windows Build Issues

FurryAcetylCoA:
This PR fix several compiling errors on Windows ( both Cygwin and MSYS2)

Since MSVC supports %lld and int64_t starting from Visual Studio 2015, the use of __int64 in headers is no longer necessary.

Note: libglib in Cygwin is unmaintained and is not fully compatible with GCC version 11 or later.

### DIFF
--- a/lib/libgtkwave/src/gw-fst-loader.c
+++ b/lib/libgtkwave/src/gw-fst-loader.c
@@ -753,7 +753,7 @@ static GwDumpFile *gw_fst_loader_load(GwLoader *loader, const char *fname, GErro
     self->subvar_jrb = make_jrb(); /* only used for attributes such as generated in VHDL, etc. */
     self->synclock_jrb = make_jrb(); /* only used for synthetic clocks */
 
-    uint numfacs = fstReaderGetVarCount(self->fst_reader);
+    uint64_t numfacs = fstReaderGetVarCount(self->fst_reader);
 
     GwFacs *facs = gw_facs_new(numfacs);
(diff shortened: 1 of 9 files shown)


lib/libgtkwave/src/gw-fst-loader.c
```diff
@@ -753,15 +753,15 @@ static GwDumpFile *gw_fst_loader_load(GwLoader *loader, const char *fname, GErro
     self->subvar_jrb = make_jrb(); /* only used for attributes such as generated in VHDL, etc. */
     self->synclock_jrb = make_jrb(); /* only used for synthetic clocks */
 
-    uint numfacs = fstReaderGetVarCount(self->fst_reader);
+    uint64_t numfacs = fstReaderGetVarCount(self->fst_reader);
```
rfuest, line 756: GTKWave and libgtkwave use the integer types that are provided by [GLib](https://docs.gtk.org/glib/types.html) instead of the `stdint.h` ones:
```suggestion
    guint64 numfacs = fstReaderGetVarCount(self->fst_reader);
```
This doesn't apply to the other libraries, like `libfst` which don't depend on GLib/GTK.

rfuest (changes requested):
>Note: libglib in Cygwin is unmaintained and is not fully compatible with GCC version 11 or later.

I think we should focus on MSYS2 and MSVC for Windows. There is also no GTK 4 package for Cygwin, which will be a problem in the future.

rfuest (approved):

PR #357, opened 2024-06-04
Unify fileselbox implementation across all platforms

FurryAcetylCoA:
The `fileselbox` used to have three different implementations for Windows(mingw), Linux and OSX.
This PR unify those three different implementations by using the `GtkFileChooserNative` API.

The "Custom Filter" filter has also been removed for the following reasons:
1. It was added back in GTK1 era. I am not sure how the `FileChooserDialog` looked back then, but I suppose there was no built-in search so the "Custom Filter" was necessary. Nowadays GTK provide built-in search in both `FileChooserDialog` and `FileChooserNative`, so user can use that instead of "Custom filter"
2. `FileDialog` API will replace both `FileChooserDialog` and `FileChooserNative` in GTK4.10. Which doesn't allow adding custom widget to the dialog window.

Additionally, migrating to `FileDialog` from `FileChooserNative` is easier since their methods are more similar. Another benefit of `FileChooserNative` is that Linux users can use their native file chooser with `GTK_USE_PORTAL=1`.

This PR also fix the issue where "gtkwave -n [FOLDER]" does not respond even when the dumpfile is properly provided. (however, it seems nobody has used "gtkwave -n [FOLDER]" for a long time since it use `fileselbox_old` which doesn't work on any platform except windows(mingw))

This PR also remove unused bridge codes in cocoa_misc. I plan to remove all of it in the future.

Tested on Windows(mingw), Windows(cygwin), Linux(x11-kde6) and OSX(latest). 

Reading any file
Reading .tcl file
Writing .gtkw file


rfuest:
I haven't had time to try this yet, but using the newer file selection API is a good idea.

I really appreciate all the help with GTKWave recently, but I would suggest that you don't spend too much time on UI changes without discussing them first. My plan for porting GTKWave to GTK 4 has changed and I'm now planning to start the UI from scratch, instead of porting the existing code over. Modern GTK applications are written very differently from the GTK 1 era code that GTKWave still uses in many places. After a MVP version of GTKWave 4 is finished, we can reevaluate which parts of the old UI code we can be reused. I've already started to implement the GTK 4 port, but it is currently blocked by the unfinished refactoring of the dump file loaders.

rfuest (approved):
I've tweaked the changelog entry a bit, but the rest looks good.

docs/man/gtkwave.1.md
```diff
@@ -30,7 +30,7 @@ use IBM EDA toolsets.
 
 **-n**,**\--nocli** \<*directory name*\>
 
-:   Use file requester for dumpfile name.
+:   Use file selection dialog for dumpfile or savefile.
```
rfuest, line 33: Just a heads up: I noticed that you probably changed this file manually, because it differs in the dot at the end of the line. You can use the `generate_man_pages.sh` script to rebuild the markdown version of the man pages from the files in the `man` directory. But I've also forgotten to regenerate the docs after the recent changes to the man pages, which I'll fix soon.

rfuest: But now that I think about that some more, maybe we shouldn't do that at all and instead regenerate the markdown from the LTS version, which the docs target at the moment. Maybe it wasn't such a good idea to put the docs directory in the same repo as the development version of GTKWave.

FurryAcetylCoA:
> I haven't had time to try this yet, but using the newer file selection API is a good idea.
> 
> I really appreciate all the help with GTKWave recently, but I would suggest that you don't spend too much time on UI changes without discussing them first. My plan for porting GTKWave to GTK 4 has changed and I'm now planning to start the UI from scratch, instead of porting the existing code over. Modern GTK applications are written very differently from the GTK 1 era code that GTKWave still uses in many places. After a MVP version of GTKWave 4 is finished, we can reevaluate which parts of the old UI code we can be reused. I've already started to implement the GTK 4 port, but it is currently blocked by the unfinished refactoring of the dump file loaders.

I agree. Much of the legacy code in GTKWave, particularly in the UI section, was tailored for a specific version of GTK on a specific platform a long time ago. These solutions no longer suits modern GTK4 applications(or even gtk3)

Rather than making piecemeal amendments that could potentially complicate the codebase further, I believe separating the core functions from the UI is indeed a more strategic approach
I appreciate your work on libgtkwave.
Also, I agree that discussing potential changes before implementation is a wise practice. Thanks for your advice. 
Please let me know if there are any specific areas where you'd like my input or assistance as you progress with the GTK4 port.

docs/man/gtkwave.1.md
FurryAcetylCoA: Oh, that's right. I forgot the docs should reflect the LTS version of GTKWave, not the experimental 3.4 version.

rfuest:
> Please let me know if there are any specific areas where you'd like my input or assistance as you progress with the GTK4 port.

Thanks for offering your help. Would you be interested in starting to port RTLBrowse over to GTK4? I find it hard to split the work on the main GTKWave application in its current state, but RTLBrowse is separate and also needs to be ported at some time. A GTK4 port of RTLBrowse could use [`GtkSourceView`](https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/) to display the source code with nice syntax highlighting.

FurryAcetylCoA:
> Thanks for offering your help. Would you be interested in starting to port RTLBrowse over to GTK4? I find it hard to split the work on the main GTKWave application in its current state, but RTLBrowse is separate and also needs to be ported at some time. A GTK4 port of RTLBrowse could use [`GtkSourceView`](https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/) to display the source code with nice syntax highlighting.

Thanks for your advice. I will look into it  this week.

PR #362, opened 2024-06-13
Fix undefined behaviour in fstMmap2 on 32-bit Windows

nickg:
This is a mistake introduced by my earlier PR #191. On 32-bit Windows `size_t` is a 32-bit type so shifting by 32 or more places is undefined behaviour. Avoid this by casting to `DWORD64` first.

### DIFF
--- a/lib/libfst/fstapi.c
+++ b/lib/libfst/fstapi.c
@@ -347,8 +347,9 @@ return(NULL);
 
 static void *fstMmap2(size_t __len, int __fd, fst_off_t __off)
 {
+DWORD64 len64 = __len;  /* Must be 64-bit for shift below */
 HANDLE handle = CreateFileMapping((HANDLE)_get_osfhandle(__fd), NULL,
-				  PAGE_READWRITE, (DWORD)(__len >> 32),
+				  PAGE_READWRITE, (DWORD)(len64 >> 32),
 				  (DWORD)__len, NULL);
 if (!handle) { return NULL; }
 


rfuest (changes requested):
I think it would be a good idea to add a comment to document the reason why this cast is used to make sure it isn't accidentally removed later on.
IMO this is a bit hard to decipher: `(DWORD)((DWORD64)__len >> 32)`, adding a temporary `DWORD64 len64` could make this easier to read without having to think about the operator precedence to understand the order of operations (cast and shift).

tbybell:
I myself would normally code it pretty much exactly like Nick did, but unfortunately, compilers are getting far more annoying with respect to warnings in the vicinity of casting because they expect user mistakes.  Farther down the file, I had to break up some crazy casting into individual assigns or gcc at the super warning level didn't like it.  That said, assigning to a DWORD64 temp probably is in order for the mmap replacement just so we're not updating it again later.
```

if(vm4ip[2] != prev_alias)
  {
  int32_t  t_i32 = ((int32_t)(prev_alias = vm4ip[2])); /* vm4ip is generic unsigned data */
  int64_t  t_i64 = (int64_t)t_i32; /* convert to signed */
  uint64_t t_u64 = (uint64_t)t_i64; /* sign extended through 64b will be logically shifted */

  fpos += fstWriterSVarint(f, (int64_t)((t_u64 << 1) | 1));
  /* all in this block was: fpos += fstWriterSVarint(f, (((int64_t)((int32_t)(prev_alias = vm4ip[2]))) << 1) | 1); */
  }
```

rfuest (approved):

PR #363, opened 2024-06-14
Correct the type of yyleng

### LINKED ISSUE #361: Abnormal memory usage in rtlbrowse
When using rtlbrowse to browse smaller designs, such as `example/des.v`, and navigate though the tree selector. The virtual memory usage spikes to over 1000G before returning to a normal level. 
For larger designs, this spike can exceed 100T and the rtlbrowse just freeze.

It appears this isn't a classic memory leak. Given that in the first scenario where memory usage does return to a normal level after a few seconds.

Also, this issue doesn't seem to exist in gtkwave 3.3 lts 

I'm looking into this further and will post updates here. 

By the way, have there been any changes or updates in GTKWave that might affect its interaction with rtlbrowse?

FurryAcetylCoA:
According to [The Open Group Base Specifications Issue 7](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/lex.html)

> The external **int** yyleng shall be set to the length of the matching string.

Close #361


rfuest:
Great catch and according to the standard this would be the correct solution, but unfortunately different `flex` versions seem to use different types for the `yyleng` variable (see https://github.com/westes/flex/issues/118 and https://github.com/opencog/link-grammar/issues/1509). TLDR of the two issues is that newer `flex` versions now use the correct `int` type, but on Apple systems it might still be `size_t`. 

I think we should let `flex` generate a header file, which will always contain a declaration of `yyleng` with the correct type.

tbybell:
> TLDR of the two issues is that newer `flex` versions now use the correct `int` type, but on Apple systems it might still be `size_t`.
> 
> I think we should let `flex` generate a header file, which will always contain a declaration of `yyleng` with the correct type.

I'm using Centos7 on my one machine and it's still size_t.  Sure, I know...it's supposed to be EOL in two weeks, but systems that use 2.5.x of flex are definitely out there.

I agree about using --header-file=whatever.h as an argument to flex.  Nobody is using the old AT&T lex anymore which was a highly annoying compatibility issue back in the old days when this project started, and going back even farther to the structural verilog compiler/simulator.

FurryAcetylCoA:
The new implementation use header file generated by flex. So it should settle the discrepancy of `yyleng` on different version of `flex` across different platform.

----
Details (from commit log):

This commit mainly aims at correcting the type of yyleng across Linux and OSX. Since the flex on those two platforms uses different types for yyleng. tbybell also mentioned that flex 2.5 on centos use size_t.

To address this discrepancy, rather than relying on manually written extern declarations, we now use the header file (`vlex.yy.h`) generated by `flex`.

flex_gen will generate both `vlex.yy.c` and `vlex.yy.h`. Notice: `vlex.h` is handwritten and still required by `vlex.l`

Additionally, this commit relocates additional `flex` options from `meson.build` to `vlex.l`. Actually, extra_args must be added to arguments via @EXTRA_ARGS@ explicitly, so the previous version of the build file didn't pass those flags to flex at all

rfuest (approved):
LGTM

PR #365, opened 2024-06-16
Rtlbrowser refactoring and migration to gtk4

FurryAcetylCoA:
This PR aims to migrate Rtlbrowser to GTK4. As discussed in https://github.com/gtkwave/gtkwave/pull/357#issuecomment-2165727108 .

Roadmaps:
- [x] Basic code cleanup in startup part
- [x] Replace gtk_main() and related APIs with GtkApplication
  - [x] Use GLib to handle the command-line argument
- [x] Apply "Preparation in GTK 3.x" part from [migration guide](https://docs.gtk.org/gtk4/migrating-3to4.html)
  - [x] Replace "key_press_event" with "key-pressed"
  - [x] Remove usage of "scroll_event"
  - [x] Replace GtkToolbar with GtkBox
  - [x] Use "button_set_image" instead of "container_add" for adding icon to button
  - [x] Replace "container_add" with "paned_add1" for adding child to GtkPaned
  - [x] Replace "notebook_detach_tab" with "widget_destroy" when removing a tab
- [x] Conduct initial testing on other platforms
  - [x] Windows (mingw64) (Also see: #367)
  - [x] Linux (Ubuntu 24.04 & Manjaro)
  - [x] OSX
- [x] Change build config so Rtlbrowser is build against gtk4 
- [x] Apply "Changes that need to be done at the time of the switch" part from [migration guide](https://docs.gtk.org/gtk4/migrating-3to4.html)
  - [x] Api changes
  - [x] Remove DnD support (Temporarily)
  - [x] Replace GtkTreeModel/GtkTreeStore with  GListModel/GtkTreeListModel
    - [x] Create a GObject to store those data 
  - [x] Update CI environment setup
- [x] Test on other platforms
- [ ]  ~Use GtkSourceView to display the source code~
  - [ ] ~Understand the source view mechanism in the previous version~
  - [ ] ~Implement syntax highlighting~

Additional Notes:
* This PR will not encompass significant UI layout changes. 
* The current IPC method using shared memory will also remain intact.


rfuest:
That looks really good, great work so far! Let me know if you need any help.

FurryAcetylCoA:
I noticed that rtlbrowse supports drag and drop functionality. However, it appears to be exclusively for Tcl scripts. There is no documentation explaining how to use Tcl scripts with rtlbrowse, and I have not come across anyone using rtlbrowse in this way. Given this, I am considering removing the DND code in rtlbrowse, as it seems to be a historical remnant and does not serve any current purpose.

What do you think about this? @rfuest

rfuest:
Drag and drop between different instances of GTKWave or rtlbrowse is handled by building a pseudo Tcl script that is parsed by the receiving GTKWave instance. With GTKWave LTS you can even show the generated Tcl script by dragging a trace from the SST into a terminal. In the development branch GTKWave uses a custom mime type to prevent this. But none of this code relies on the Tcl interpreter, it just uses Tcl as a serialization format.

The code to manually serialize and deserialize to and from Tcl is very complex and I would like to replace it in GTKWave 4. I'm planning to use JSON (like for the save format) or GVariant (which is internally used by D-Bus) as the serialization format.

If it makes the refactoring easier you can go ahead and remove the Tcl based DnD code. Just mark it as a current limitation in the Changelog or open an issue to track this limitation, after this PR is merged.

tbybell:
I don't remember the exact details except for encapulation into { } marked data items for that pseudo-Tcl script, but the way it works (and the way I originally tested it) was that you can drag and drop from rtlbrowse into an X11 version of emacs (!) to see what the data format is to communicate between rtlbrowse and gtkwave.

It still works.  Try it.

The {net ...} syntax for the data items came from user submissions for another tool:

External DnD support is Copyright (C) 2008-2024 Concept Engineering GmbH.

https://www.concept.de/RTLvision.html
...take a good hard look at the wave viewer in some of the screenshots.  It's almost unrecognizable now, but you can spot some lingering historical similarities that remain.  :-)

FurryAcetylCoA:
Most of the migration work is complete except for the changes to GtkSourceView.

Please take some time to review and test the changes and let me know if you find any bugs, functional inconsistencies with the previous version, suggestions, code style issues, etc.

contrib/rtlbrowse/logfile.c
```diff
@@ -93,1390 +89,1308 @@ static GtkTextTag *blue_tag = NULL, *fwht_tag = NULL;
...
+    gchar *line_text, *found, *pnt;
+    gint offset;
+    gchar *strcaseins;
+
+    if (!str)
+        return (FALSE);
+
+    pnt = strcaseins = strdup(str);
+    while (*pnt) {
+        *pnt = toupper((int)(unsigned char)*pnt);
```
rfuest, line 240: This is outside the scope of this PR, but in the future we need to decide how text encodings will be handled by rtlbrowse and GTKWave in general. This function assumes that the input is UTF-8 encoded below by using `g_utf8_strlen`, but also uses `toupper` to convert to uppercase, which might or might not use UTF-8.

To fix this properly we need to get rid of the `WAVE_LOCALE_FIX`, which already causes encoding issues when `rtlbrowse --help` tries to displays some localized german help text with umlauts.

contrib/rtlbrowse/logfile.c
```diff
@@ -93,1390 +89,1308 @@ static GtkTextTag *blue_tag = NULL, *fwht_tag = NULL;
...
-		{
-		gtk_text_buffer_select_range(tb, &match_start, &match_end);
-		read_insert_position(tr);
-		tr->srch_line = tr->line;
-		tr->srch_offs = tr->offs;
-
-		/* tm = gtk_text_buffer_get_insert(tb); */ /* scan-build : never read */
+    struct text_find_t *tr = selected_text_via_tab;
+
+    if ((tr) && (tr->text)) {
```
rfuest, line 287: These kind of checks are common in the GTKWave code base. By negating them it is possible to reduce the nesting depth and make the code more readable:
```suggestion
    if (tr == NULL || tr->text == NULL) {
        return;
    }
```

contrib/rtlbrowse/logfile.c
```diff
@@ -93,1390 +89,1308 @@ static GtkTextTag *blue_tag = NULL, *fwht_tag = NULL;
...
-    gtk_box_pack_start (GTK_BOX (hbox), matchcase_checkbutton, FALSE, FALSE, 0);
-    }
-
+    gtk_check_button_set_active(GTK_CHECK_BUTTON(matchcase_checkbutton), matchcase_active);
+    g_signal_connect(matchcase_checkbutton, "toggled", G_CALLBACK(toggle_callback), NULL);
 
+    gtk_box_append(GTK_BOX(hbox), matchcase_checkbutton);
+}
 
 static char *tmpnam_rtlbrowse(char *s, int *fd)
```
rfuest, line 548: `tmpnam_rltbrowse` was used by drag and drop, but I don't think we'll use it in the future and this can be removed.

contrib/rtlbrowse/logfile.c
```diff
@@ -93,1390 +89,1308 @@ static GtkTextTag *blue_tag = NULL, *fwht_tag = NULL;
...
-
-	jrb_free_tree(strs);
-	g_free(sel);
-	}
-
-}
-
-
-
+#define WAVE_DRAG_TAR_NAME_0 "text/plain"
```
rfuest, line 712: Let's just remove the drag and drop code entirely, we can always refer back to the GIT history or LTS version if we need to.

contrib/rtlbrowse/logfile.c
```diff
@@ -93,1390 +89,1308 @@ static GtkTextTag *blue_tag = NULL, *fwht_tag = NULL;
...
 
-if (gtk_text_buffer_get_selection_bounds (gtk_text_view_get_buffer(GTK_TEXT_VIEW(text)),
-                                         &start, &end))
-       {
-       if(gtk_text_iter_compare (&start, &end) < 0)
-               {
-               sel = gtk_text_buffer_get_text(gtk_text_view_get_buffer(GTK_TEXT_VIEW(text)),
-                                              &start, &end, FALSE);
+    s2 = malloc(strlen(t->ctx->which->fullname) + 1 + strlen(s) + 1);
+    sprintf(s2, "%s.%s", t->ctx->which->fullname, s);
```
rfuest, line 1062:
```suggestion
    s2 = g_strdup_printf("%s.%s", t->ctx->which->fullname, s);
```

contrib/rtlbrowse/logfile.c
```diff
@@ -93,1390 +89,1308 @@ static GtkTextTag *blue_tag = NULL, *fwht_tag = NULL;
...
+        if (!strcmp(s2, ft->fullname)) {
+            bwlogbox(ft->fullname, 640 + 8 * 8, ft, 0);
+            break;
+        }
 
+        ft = ft->next_flat;
+    }
 
-return(FALSE); /* call remaining handlers... */
+    free(s2);
```
rfuest, line 1073: To match the change above:
```suggestion
    g_free(s2);
```

contrib/rtlbrowse/logfile.c
```diff
@@ -93,1390 +89,1308 @@ static GtkTextTag *blue_tag = NULL, *fwht_tag = NULL;
...
+    (void)x; (void)y;
+    gchar *sel;
+
+    GtkTextIter start;
+    GtkTextIter end;
+
+    if (gtk_text_buffer_get_selection_bounds(gtk_text_view_get_buffer(GTK_TEXT_VIEW(text)),
+                                             &start,
+                                             &end)) {
+        if (gtk_text_iter_compare(&start, &end) < 0) {
```
rfuest, line 1095: This is another case where the nesting could be reduced to improve readability:
```suggestion
    if (!gtk_text_buffer_get_selection_bounds(gtk_text_view_get_buffer(GTK_TEXT_VIEW(text)),
                                             &start,
                                             &end)) {
        return;
    }

    if (gtk_text_iter_compare(&start, &end) >= 0) {
        return;    
    }
```

contrib/rtlbrowse/logfile.c
```diff
@@ -93,1390 +89,1308 @@ static GtkTextTag *blue_tag = NULL, *fwht_tag = NULL;
...
+                                   GTK_POLICY_AUTOMATIC,
+                                   GTK_POLICY_AUTOMATIC);
+    gtk_scrolled_window_set_child(GTK_SCROLLED_WINDOW(scrolled_window), text);
 
-g_signal_connect(text, "button_release_event",
-                       G_CALLBACK(button_release_event), NULL);
+    gtk_widget_set_margin_start (scrolled_window, 5);
+    gtk_widget_set_margin_end   (scrolled_window, 5);
+    gtk_widget_set_margin_top   (scrolled_window, 5);
+    gtk_widget_set_margin_bottom (scrolled_window, 5);
```
rfuest, line 1235: The margin should be applied to the text view and not the scrolled window.

rfuest (changes requested):
In my first brief test I didn't find anything that didn't work as good as the old `rtlbrowse` version did. Nice work.

I did a first review pass, where I tried to focus on changes you made and not issues with the existing code base. But in some cases the suggested changes might be outside the scope of this PR, if the problems were already present in the code. If this is the case we can just start to create issues to track improvements that could be made to `rtlbrowse` in the future.

Could you add the `des.stems` file for the example to the repo to make it easier to try `rtlbrowse` in the future?

contrib/rtlbrowse/main.c
```diff
@@ -0,0 +1,188 @@
...
+    int i;
+    int len;
+
+    len = strlen(arg);
+
+    /* Determine whether arg is file name or shmid */
+    for (i = 0; i < len; i++) {
+        if (!g_ascii_isxdigit(arg[i]))
+            break;
+    }
```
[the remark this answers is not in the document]
FurryAcetylCoA: solved in c2b5d7f3

contrib/rtlbrowse/main.c
```diff
@@ -0,0 +1,188 @@
...
+{
+    GVariantDict *options;
+    gchar **cmd;
+
+    options = g_application_command_line_get_options_dict(cmdline);
+    if (!g_variant_dict_lookup(options, G_OPTION_REMAINING, "^a&s", &cmd) ||
+        g_strv_length(cmd) != 1) {
+        // g_option_context_get_help can not be used to render help text here.
+        // Since the `context` it need is constructed inside g_application_parse_command_line,
+        // using private members of GApplication. Is there another way to print the help?
```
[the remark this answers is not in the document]
FurryAcetylCoA: solved in c2b5d7f3
  
`GApplication::open` is not utilized in this version yet. We may(or have to?) implement it in the future version when user can launch rtlbrowse without any command line argument and select file from the GUI. So `g_application_open()` could be used together.

contrib/rtlbrowse/logfile.c
```diff
@@ -93,1390 +89,1308 @@ static GtkTextTag *blue_tag = NULL, *fwht_tag = NULL;
...
+// gpointer user_data)
+// {
+//     pressWindow = gtk_event_controller_get_widget(GTK_EVENT_CONTROLLER(self));
+//
+//     pressX = x;
+//     pressY = y;
+//
+// }
+
+static gint expose_event_local(GtkWidget *widget)
```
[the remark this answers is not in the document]
FurryAcetylCoA: solved in f70a7043

contrib/rtlbrowse/logfile.c
FurryAcetylCoA: solved in 76abb3c2

contrib/rtlbrowse/logfile.c
FurryAcetylCoA: solved in f70a7043

FurryAcetylCoA:
The source view and annotation mechanism is much more complicated than i expected. 

The `bwlogbox_2`  implements parsing, rendering, variable name extraction, and wave dump interaction. It is the core of rtlbrowse. 
It uses the `log_text*` API family to render the rich text inside the text view, token by token. This means that we can technically add more syntax highlighting directly into it.

I am currently preoccupied with NPEE, hence I don't have enough time to thoroughly comprehend it and ensure behavior consistency after migration to `GtkSourceView`. Additionally, I haven't figure out how to add values into the view. (Does `GtkSourceView` provides something like ghost text or inline annotation?). Therefore, this PR will __not__ include migration to `GtkSourceView`, sorry for that.

By the way, it appears that `rtlbrowse` does not(or cannot) distinguish between data at the Q and D terminals of register types. It could potentially confuse user when such register is used as a condition in HDL.

rfuest:
I would like to merge this with a rebase to keep your commit messages in the GIT history. But I think we should a "rtlbrowse:" prefix to the commit messages to distinguish them from other changes.

Don't worry about `GtkSourceView`, that was just an idea I had for a possible improvement. The state `rtlbrowse` is in now is a good starting point for future improvements and it's easier to work on incremental improvements in smaller PRs.

FurryAcetylCoA:
All commit directly related to rtlbrowse is now prefixed with `rtlbrowse: `

rfuest (approved):
LGTM. Thanks for the getting the migration of GTKWave to GTK 4 going!

PR #367, opened 2024-06-18
Fix: Failed to create rtlbrowse subprocess on windows

FurryAcetylCoA:
In the CreateProcess function, the first parameter `lpApplicationName` does not use the search path, as documented in the [WINAPI Doc](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa#parameters).

The second parameter `lpCommandLine` uses the search path if `lpApplicationName` is NULL.

Tested on msys running on windows 11. 

However, due to Microsoft's removal of older documentation, I am uncertain whether this change might impact previous versions of Windows. Please let me know if I should setup such environment and test it.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -2358,7 +2358,7 @@ void activate_stems_reader(char *stems_name)
 
                 update_time_box();
 
-                rc = CreateProcess("rtlbrowse.exe",
+                rc = CreateProcess(NULL,
                                    mylist,
                                    NULL,
                                    NULL,


rfuest:
> However, due to Microsoft's removal of older documentation, I am uncertain whether this change might impact previous versions of Windows. Please let me know if I should setup such environment and test it.

Not, that's not necessary. It should work the same on old versions of Windows and the docs don't seem to have changed much in the last 16 years: https://web.archive.org/web/20080508195209/http://msdn.microsoft.com/en-us/library/ms682425(VS.85).aspx 
I'm not sure which versions of Windows are still officially supported by GTK 4, because it wouldn't make sense if we try to support older versions of Window than our GTK dependency allows.

rfuest (approved):

PR #381, opened 2024-09-07
Fix Ubuntu 24.04/clang 18.0 warning on chk_len

wsnyder:
This fixes Clang 18 (default clang version in Ubuntu 24.04) compiler errors.

```
gtkwave/fstapi.c:554:19: error: variable length arrays in C++ are a Clang extension [-Werror,-Wvla-cxx-extension]
  554 | unsigned char buf[chk_len];
      |                   ^~~~~~~
gtkwave/fstapi.c:554:19: note: read of non-const variable 'chk_len' is not allowed in a constant expression
gtkwave/fstapi.c:553:5: note: declared here
  553 | int chk_len = 5; /* TALOS-2023-1783 */
      |     ^
```

### DIFF
--- a/lib/libfst/fstapi.c
+++ b/lib/libfst/fstapi.c
@@ -550,8 +550,9 @@ return(rc);
 
 static uint32_t fstReaderVarint32(FILE *f)
 {
-int chk_len = 5; /* TALOS-2023-1783 */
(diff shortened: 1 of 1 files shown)


tbybell:
Thanks for the patch.

PR #387, opened 2024-10-09
Fix duplicate 'releases' tags in metainfo

pghmcfc:
They cause metainfo validation to fail.

Looks like a copy-and-paste issue.


rfuest:
Thanks!

tbybell:
We apologise for the fault in the metainfo.xml. Those responsible have been sacked.
(Monty Python and the Holy Grail reference for those who don't know)

PR #388, opened 2024-10-17
Add RISC-V target to CI flow

hanbings:
This PR adds build and test steps in CI for RISC-V Linux using `uraimo/run-on-arch-action`.
While gtkwave itself is architecture-independent. Including tests on another architecture might aids packagers on other distros. Just in case there are any niche requirements / inconsistency specific to different platforms.
Apt is utilized instead of pip to install meson and ninja because the python base environments is externally managed in the building container.

Note: This PR does not produce a distributable package for RISC-V Linux. As stated in Meson [manual](https://mesonbuild.com/Creating-Linux-binaries.html), "app is guaranteed to break as different distros have binary-incompatible C++ libraries" ( unless statically linked )

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,36 @@ jobs:
       - name: Run tests
         run: meson test -C build -v
 
+  build-linux-riscv:
+    name: Build (Linux-RV)
+    runs-on: ubuntu-latest
+    # Tighten permissions granted to the GITHUB_TOKEN
+    permissions:
+      packages: write # for uraimo/run-on-arch-action to cache docker images
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Build
+        uses: uraimo/run-on-arch-action@v2
+        with:
+          arch: riscv64
+          distro: ubuntu_latest
+          # Not mandatory, but speeds up builds by storing container images
+          #with dependencies installed in a GitHub package registry.
+          githubToken: ${{ github.token }}
+
+          # Install some dependencies in the container
+          # sudo is not required
+          install: |
+            apt-get update
+            apt-get -y install gperf desktop-file-utils libgtk-3-dev libgtk-4-dev libjudy-dev libgirepository1.0-dev
+            apt-get -y install meson ninja-build flex
+
+          run: |
+            meson setup build
+            meson compile -C build
+            meson test -C build -v 
+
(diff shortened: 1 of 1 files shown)


rfuest:
I like the idea of testing on RISC-V, but the build is a lot slower than the native builds. If we disregard the time it took to generate the docker image, which would be cached in subsequent runs, the RISC-V build took about 5.5 minutes to complete.

The fast CI runtimes did make iterating in case of a CI failure very quick and I'm not sure I want to give that up for a fairly niche architecture (at least at the moment). Our macOS CI build does at least test a non X86 architecture and a different OS.

A compromise might be to cross compile for RISC-V and only run the tests via QEMU. While this isn't quite the same as running the entire build on RISC-V, it should be a lot faster and also tests the ability to cross compile GTKWave.

> Apt is utilized instead of pip to install meson and ninja because the python base environments is externally managed in the building container.

Interesting, this means that the RISC-V build did run on Ubuntu 24.04 LTS and not on 22.04 like the other tests. pip was only used because the packaged meson version in 22.04 was too old.

> Note: This PR does not produce a distributable package for RISC-V Linux. As stated in Meson [manual](https://mesonbuild.com/Creating-Linux-binaries.html), "app is guaranteed to break as different distros have binary-incompatible C++ libraries" ( unless statically linked

Creating nightly packages in CI is a long time goal, not only for RISC-V. For Linux the best option, in my opinion, would be to build Flatpaks, which don't have the issue of having to deal with ABI incompatible distros.

hanbings:
Thank you for the detailed advice.

While Meson supports setting up cross toolchains out of the box, cross compiling GTKWave requires guest architecture library files (cross-dependencies). 
We could use cyberjunk/gha-ubuntu-cross to download and install apt packages for another architecture (essentially add another architecture to sources.list.d)

However the Debian wiki explicitly stated that programs require gobject-introspection to compile cannot be cross compiled at this point for [reasons](https://wiki.debian.org/CrossBuildPackagingGuidelines#gobject-introspection)     // and [more](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1030223)

Unfortunately, libgtk-4-dev:riscv64 depends on gobject-introspection:riscv64 which is currently impossible to install on an amd64 building machine.
In conclusion, GTKWave can't be cross compiled for RISC-V at this time, and I will therefore close this PR.

PR #398, opened 2025-01-07
fix left-shift and signedness warnings

joergen7:
This pr fixes an issue where a negative number is left-shifted in ghw, an implicit cast from a signed to an unsigned data type in vzt and improves upon an earlier fix that introduced a limit in the size of an id string in fst.

## ghw: left-shift of negative number

The left-shift of a negative number causes a warning in ghw.

### Observation

Compiling gtkwave produces the following warning:

    ../lib/libghw/libghw.c:276:25: warning: left shift of negative value [-Wshift-negative-value]
      276 |             r |= r_mask << off;
          |                  ~~~~~~~^~~~~~

### Execution

The fix creates the mask not by shifting in zeros into a one-vector from the right but by shifting in ones into a zero-vector from the right and then negating.

## vzt: implicit cast from signed to unsigned

An implicit cast from a signed char pointer to an unsigned char pointer causes a warning in vzt.

### Observation

Compiling gtkwave produces the following warning:

    ../lib/libvzt/vzt_read.c:317:22: warning: pointer targets in initialization of ‘unsigned char *’ from ‘char *’ differ in signedness [-Wpointer-sign]
      317 | unsigned char *pnt = b->mem;
          |                      ^
    ../lib/libvzt/vzt_read.c:405:5: warning: pointer targets in assignment from ‘char *’ to ‘unsigned char *’ differ in signedness [-Wpointer-sign]
      405 | pnt = (char *)(val_dict + (num_dict_words = num_dict_entries * num_sections));
          |     ^
    ../lib/libvzt/vzt_read.c:434:5: warning: pointer targets in assignment from ‘char *’ to ‘unsigned char *’ differ in signedness [-Wpointer-sign]
      434 | pnt = (char *)(b->vindex + num_bitplanes * lt->total_values);
          |     ^
    ../lib/libvzt/vzt_read.c:442:30: warning: pointer targets in assignment from ‘unsigned char *’ to ‘char *’ differ in signedness [-Wpointer-sign]
      442 |                 b->sindex[i] = pnt;
          |                              ^
    ../lib/libvzt/vzt_read.c:443:32: warning: pointer targets in passing argument 1 of ‘strlen’ differ in signedness [-Wpointer-sign]
      443 |                 pnt += (strlen(pnt) + 1);
          |                                ^~~
          |                                |
          |                                unsigned char *

### Execution

I declared `pnt` to be signed. Not only mutes this the warning from the implicit cast when `pnt` is assigned from `b->mem` but also mutes the warnings when pnt is wrote back into `b->sindex` and when `pnt` is passed to `strlen`.

## fst: fix size limit of id string

An earlier fix that limited the size of an id string in fst (see #397) poorly enforces the required size limit of 14.

### Observation

Since the while loop might be entered, when the `len` variable is 14 initially, we end up with id strings of length up to 15.

### Execution

Enforce the `len` variable to be lower than 14 prior to entering the loop.

### Observation

`gw-vcd-file.h` includes `gtkwave.h`. Both files reside in the same directory. Therefore a quoted include is preferable.

### Execution

Make include quoted.

### DIFF
--- a/lib/libfst/fstapi.c
+++ b/lib/libfst/fstapi.c
@@ -3910,7 +3910,7 @@ char *pnt = buf;
  int len = 0;
 
 /* zero is illegal for a value...it is assumed they start at one */
-while (value && len <= 14)
+while (value && len < 14)
         {
         value--;
 	++len;
--- a/lib/libghw/libghw.c
+++ b/lib/libghw/libghw.c
@@ -259,7 +259,6 @@ ghw_read_sleb128 (struct ghw_handler *h, int32_t * res)
 int
 ghw_read_lsleb128 (struct ghw_handler *h, int64_t * res)
 {
-  static const int64_t r_mask = -1;
   int64_t r = 0;
   unsigned int off = 0;
 
@@ -273,7 +272,7 @@ ghw_read_lsleb128 (struct ghw_handler *h, int64_t * res)
       if ((v & 0x80) == 0)
 	{
 	  if ((v & 0x40) && off < 64)
-	    r |= r_mask << off;
+	    r |= ~((1 << off)-1);
 	  break;
 	}
     }
--- a/lib/libghw/libghw.c
+++ b/lib/libghw/libghw.c
@@ -259,6 +259,7 @@ ghw_read_sleb128 (struct ghw_handler *h, int32_t * res)
 int
 ghw_read_lsleb128 (struct ghw_handler *h, int64_t * res)
 {
+  static const int64_t r_mask = -1;
   int64_t r = 0;
   unsigned int off = 0;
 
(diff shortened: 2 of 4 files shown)


joergen7:
Hello, would it be sufficient to revert `libghw` to make this pr viable?

joergen7:
I reverted `libghw`.

rfuest (approved):
> Hello, would it be sufficient to revert `libghw` to make this pr viable?

Sure, it wasn't my intention to make it sound like the PR wouldn't be viable to be merged. I just wanted to make sure that we don't start our own fork of `libghw`. When you decide to report the `libghw` issue upstream it would be great if you could create a new PR to sync our copy when the upstream version has been updated.

The other changes look good to me, but I'll let @tbybell take another look because he's much more familiar with the FST and VZT code.

PR #401, opened 2025-01-10
feat: Add macOS application wrapper

Emin017:
Add macOS application wrapper for gtkwave

The benefits of this change include:
1. Improved user experience for macOS users, as they can now launch gktwave directly from the Applications folder like any other native macOS app.
2. Easier distribution and installation of gktwave on macOS systems, as users can simply download and drag the application bundle to their Applications folder.


Emin017:
Some previews:
图片
</br>
图片

rfuest:
Thanks for the PR. Any help to improve the experience on macOS is always appreciated.

Does this only build the wrapper with Xcode or does it compile GTKWave entirely inside the Xcode project? I don't like the idea of having a second build system that needs to be kept in sync with the `meson` based build system. It seems to be possible to create macOS apps from meson, but I don't know how complicated that would be to setup: https://mesonbuild.com/Creating-OSX-packages.html

> 2\. Easier distribution and installation of gktwave on macOS systems, as users can simply download and drag the application bundle to their Applications folder.

Would it make sense to make the application that is built by CI available to users as a build artefact? Or is the macOS security in newer version so strict that providing an unsigned app doesn't make any sense?

macos/gtkwave/Assets.xcassets/AppIcon.appiconset/gtkwave-1024x1024.png
rfuest, line 1: The icons weren't properly scaled. The stroke width is constant in all sizes, which causes the green trace outline in this version to be too thin and too thick in the 64x64 version. Let me know if you need help with this, so that I can provide you properly scaled PNGs.

rfuest (changes requested):

Emin017:
> Does this only build the wrapper with Xcode or does it compile GTKWave entirely inside the Xcode project? I don't like the idea of having a second build system that needs to be kept in sync with the meson based build system. It seems to be possible to create macOS apps from meson, but I don't know how complicated that would be to setup: https://mesonbuild.com/Creating-OSX-packages.html

Sorry for I didn't look closely at the meson documentation, the current macos application build is inspired by [ghostty](https://github.com/ghostty-org/ghostty/tree/main/macos). I think it's a good idea to integrate the process of building macOS Application into meson, and I'll try it this way later.

> Would it make sense to make the application that is built by CI available to users as a build artefact? Or is the macOS security in newer version so strict that providing an unsigned app doesn't make any sense?

The unsigned application will triggers a Gatekeeper warning, which need additional step to run the application.
But given the [complex process and application limitations](https://developer.apple.com/support/compare-memberships/) of signing certificates, I think it makes sense to provide an application through ci (Honestly I don't think the macOS restrictions are so strict yet).

Emin017:
Superseded by #402

Emin017:
Superseded by https://github.com/gtkwave/gtkwave/pull/402

PR #402, opened 2025-01-11
Meson macos app

Emin017:
Build macOS application bundle in meson build system.

The benefits of this change include:

1. Improved user experience for macOS users, as they can now launch gktwave directly from the Applications folder like any other native macOS app.
2. Easier distribution and installation of gktwave on macOS systems, as users can simply download and drag the application bundle to their Applications folder.


lib/libfst/meson.build
```diff
@@ -16,6 +16,8 @@ libfst = static_library(
     sources: libfst_sources,
     dependencies: libfst_dependencies,
     include_directories: config_inc,
+    install: true,
+    install_dir: install_libs_rpath,
```
rfuest, line 20: Is this required? We are statically linking against all libraries except `libgtkwave`.

share/icons/meson.build
```diff
@@ -1,4 +1,4 @@
-icons_dir = get_option('datadir') / 'icons/hicolor'
+icons_dir = datadir_path / 'icons/hicolor'
```
rfuest, line 1: Why did you replace `get_option('datadir')` with `datadir_path` everywhere? `datadir_path` seems to always be equal to `get_option('datadir')`.

meson_options.txt
```diff
@@ -67,3 +67,10 @@ option(
     value: 'auto',
     description: 'Set rpath for executables (automatically enabled if the prefix is not /usr)',
 )
+
+option(
+    'build_macos_app',
+    type: 'feature',
+    value: 'disabled',
```
rfuest, line 74: This option isn't auto detected, so that a boolean is more appropriate:
```suggestion
    type: 'boolean',
    value: false,
```

meson.build
```diff
@@ -205,7 +206,18 @@ endforeach
...
+        app_path = get_option('prefix')
+        app_contents = app_path / 'Contents'
+
+        install_rpath = app_path
+        install_libs_rpath = app_contents / 'Resources' / 'lib'
+        install_headers_rpath = install_libs_rpath
+    else
+        install_rpath = get_option('prefix') / get_option('libdir')
+        install_libs_rpath = install_rpath
+        install_headers_rpath = install_rpath
```
rfuest, line 219: I don't think there is such a thing as a `rpath` for header files and I would like to keep this section closer to the original, if possible. Couldn't we also fix the library path as part of the `meson setup` call? Something like: `meson setup --libdir=Contents/Resources/lib ...`.

meson.build
```diff
@@ -231,3 +243,30 @@ gnome.post_install(
...
+
+    install_data(
+        'macos/AppIcon.icns',
+        install_dir: app_resources
+    )
+
+    install_data(
+        'macos/gtkwave.entitlements',
+        install_dir: app_contents
+    )
```
rfuest, line 271: Please rename the `app_...` variables in something like `macos_...`. All variables in meson are global and that will future variable name clashes less likely.

man/meson.build
```diff
@@ -15,4 +15,5 @@ install_man(
     'vztminer.1',
     'fstminer.1',
     'xml2stems.1',
+    install_dir: datadir_path / 'man'
```
rfuest, line 18: In the default config the `mandir` can be set independently of the `datadir`. Did you check if the man pages are even available when they are installed via the app bundle? Otherwise I would suggest to not install them for app builds.

docs/install/mac.md
```diff
@@ -72,6 +72,24 @@ GTKWave is now installed on your macOS. You can start it in your terminal, run:
...
+Ensure you have installed the dependencies as described [above](#installing-dependencies).
+Then run the following command to build the macOS application:
+```shell
+meson setup build \
+    -Dbuild_macos_app=enabled \
+    -Dset_rpath=enabled \
+    -Dintrospection=false \
+    --prefix=/Applications/GTKWave.app \
+    --datadir=/Applications/GTKWave.app/Contents/Resources/share \
+    --bindir=Contents/Resources/bin
```
rfuest, line 85: Should we add a shell script for this? I think that would make it easier to build the app and we don't need to remember to update the docs each time the parameters change.

docs/install/mac.md
```diff
@@ -72,6 +72,24 @@ GTKWave is now installed on your macOS. You can start it in your terminal, run:
...
+```
+You can find the application in your Applications folder:
```
rfuest, line 87: The instructions are missing the install step: `meson install -C build`

docs/install/mac.md
```diff
@@ -72,6 +72,24 @@ GTKWave is now installed on your macOS. You can start it in your terminal, run:
 gtkwave
 ```
 
+### Building macOS Application
+Ensure you have installed the dependencies as described [above](#installing-dependencies).
```
rfuest, line 76:
```suggestion
### Building macOS Application

Ensure you have installed the dependencies as described [above](#installing-dependencies).
```

rfuest (changes requested):
Thanks for the PR, I like this approach much better than #401.

Did you try to run the application on a system where the dependencies aren't installed? This section of the meson docs makes me wonder if there are additional steps that are required to make the app work on any macOS machine without also having to install the homebrew dependencies: https://mesonbuild.com/Creating-OSX-packages.html#external-libraries

Emin017:
>Did you try to run the application on a system where the dependencies aren't installed? This section of the meson docs makes me wonder if there are additional steps that are required to make the app work on any macOS machine without also having to install the homebrew dependencies: https://mesonbuild.com/Creating-OSX-packages.html#external-libraries

I added a shell scripts for adding runtime dependencies in bundle. Although it works, I find this thing a bit trival, because these dependencies need to be added manually by developer, as the meson docs said: 

> there is no reliable automatic way to determine how each dependency should be handled

But for now it should work.

man/meson.build
Emin017: It shouldn't be retrieved by macos by default, but I referenced an older version of [gtkwave.app](https://gtkwave.sourceforge.net/gtkwave.zip), which puts the man folder in the `gtkwave.app/Contents/Resources/share` folder.
Furthermore, if we don't set the path here, the man folder will be installed under the `GTKWave.app/man` path, which is not aligned with the macOS standards.

meson.build
rfuest: > I didn't find a way to set header files path in `meson setup`. If I set the path by using `meson setup --libdir`, I can only install it under the `gtkwave.app/include`folder, but I actually want to install it under `gktwave.app/Contents/Resources/include`

Did you try `meson configure --includedir=Contents/Resources/include`?

man/meson.build
rfuest: Would removing this line and adding yet another `setup` parameter work? `meson setup --mandir=/Applications/GTKWave.app/Contents/Resources/share/man ...`. Having to specify a ton of parameters for the setup step might not be ideal, but if it is hidden behind the build script it's not too bad IMO and I would like too keep the changes to the build for one platform to a minimum.

docs/install/mac.md
```diff
@@ -72,6 +72,19 @@ GTKWave is now installed on your macOS. You can start it in your terminal, run:
 gtkwave
 ```
 
+### Building macOS Application
+
+Ensure you have installed the dependencies as described [above](#installing-dependencies).
+Then run the following command to build the macOS application:
+```shell
+./macos/build.sh
```
rfuest, line 80: I'm not familiar with how permissions are handled under macOS. Does the `meson install` require additional permissions, similar to `sudo meson intall` on Linux? It that is the case it might be better to let the build script only build the application, which can be run without elevated permissions, and then only run the install step as admin (outside the script).

Emin017: I changed the install dir from `/Applications` to `~/Applications`, in order to avoid additional permssions.

man/meson.build
Emin017: Thank you for the suggestion! That did the works! (Just like `--includedir`, it works normally as well.) And I apologize for my unfamiliarity with the usage of meson.

meson.build
```diff
@@ -205,7 +205,17 @@ endforeach
...
 if get_option('set_rpath').enabled() or (get_option('set_rpath').auto() and get_option('prefix') != '/usr')
-    install_rpath = get_option('prefix') / get_option('libdir')
+    if host_machine.system() == 'darwin' and get_option('build_macos_app') == true
+        macos_app_path = get_option('prefix')
+        macos_app_contents = macos_app_path / 'Contents'
+
+        macos_app_exe = macos_app_contents / 'MacOS'
+        macos_app_resources = macos_app_contents / 'Resources'
+
+        install_rpath = macos_app_path
```
rfuest, line 215: Are you sure this is the correct rpath? Why wouldn't this be the library directory? Maybe setting `DYLD_LIBRARY_PATH` in `launcher.sh` wouldn't be necessary if the rpath in the binary would be correct.

rfuest, line 215: These variables are primarily used in the install section below. I think we should move them there.

```suggestion
```

meson.build
```diff
@@ -231,3 +241,29 @@ gnome.post_install(
     update_desktop_database: true,
     gtk_update_icon_cache: true,
 )
+
+if host_machine.system() == 'darwin' and get_option('build_macos_app') == true
+    install_data(
```
rfuest, line 246: See comment above.

```suggestion
if host_machine.system() == 'darwin' and get_option('build_macos_app') == true
        macos_app_path = get_option('prefix')
        macos_app_contents = macos_app_path / 'Contents'

        macos_app_exe = macos_app_contents / 'MacOS'
        macos_app_resources = macos_app_contents / 'Resources'
        
        install_data(
```

rfuest (changes requested):
A few more small issues, but it's close to be ready to merge it.

> Thank you for the suggestion! That did the works! (Just like --includedir, it works normally as well.) And I apologize for my unfamiliarity with the usage of meson.

That's not a problem, I really appreciate that you work on this. Build systems are complex and I also need to look things up all the time, even after working with meson for a few years.

meson.build
Emin017: I fixed the path to `install_rpath`, but I think the `DYLD_LIBRARY_PATH` environment variable is still worth to be set, this is to circumvent the program from using dynamic libraries in the system path, otherwise it would report an error:
```shell
$ ~/Applications/GTKWave.app/Contents/MacOS/gtkwave
objc[40352]: Class GdkQuartzView is implemented in both /opt/homebrew/Cellar/gtk+3/3.24.43/lib/libgdk-3.0.dylib (0x1011654e8) and /Users/Emin/Applications/GTKWave.app/Contents/Resources/Frameworks/libgdk-3.0.dylib (0x100ddd4e8). One of the two will be used. Which one is undefined.
objc[40352]: Class GdkQuartzNSWindow is implemented in both /opt/homebrew/Cellar/gtk+3/3.24.43/lib/libgdk-3.0.dylib (0x101165538) and /Users/Emin/Applications/GTKWave.app/Contents/Resources/Frameworks/libgdk-3.0.dylib (0x100ddd538). One of the two will be used. Which one is undefined.
objc[40352]: Class gdkCoreCursor is implemented in both /opt/homebrew/Cellar/gtk+3/3.24.43/lib/libgdk-3.0.dylib (0x101165588) and /Users/Emin/Applications/GTKWave.app/Contents/Resources/Frameworks/libgdk-3.0.dylib (0x100ddd588). One of the two will be used. Which one is undefined.

(process:40352): GLib-GObject-CRITICAL **: 09:58:14.219: cannot register existing type 'GdkPixbuf'

(process:40352): GLib-GObject-CRITICAL **: 09:58:14.220: g_type_add_interface_static: assertion 'G_TYPE_IS_INSTANTIATABLE (instance_type)' failed

(process:40352): GLib-GObject-CRITICAL **: 09:58:14.220: g_type_add_interface_static: assertion 'G_TYPE_IS_INSTANTIATABLE (instance_type)' failed

(gtkwave:40352): GLib-CRITICAL **: 09:58:14.220: g_once_init_leave_pointer: assertion 'result != 0' failed

(process:40352): GLib-GObject-CRITICAL **: 09:58:14.220: g_param_spec_object: assertion 'g_type_is_a (object_type, G_TYPE_OBJECT)' failed

(process:40352): GLib-GObject-CRITICAL **: 09:58:14.220: validate_pspec_to_install: assertion 'G_IS_PARAM_SPEC (pspec)' failed

(process:40352): GLib-GObject-CRITICAL **: 09:58:14.220: g_param_spec_ref_sink: assertion 'G_IS_PARAM_SPEC (pspec)' failed

(process:40352): GLib-GObject-CRITICAL **: 09:58:14.220: g_param_spec_unref: assertion 'G_IS_PARAM_SPEC (pspec)' failed
[1]    40352 segmentation fault  ~/Applications/GTKWave.app/Contents/MacOS/gtkwave
```
And after setting the environment variables, gtkwave works fine.

rfuest (approved):
LGTM. Thanks again!

PR #406, opened 2025-01-15
Port to gcc 15

pghmcfc:
gcc 15 strengthens the incompatible pointer type checks, particularly in respect of function pointers, where the types of arguments passed to the function are also now significant.

This change fixes one such case in `tcl_helper.h` and works around the other by using a typecast in `gtk23compat.h`, since the function argument types are not identical in all cases there.

I don't think this is likely to cause breakage with older compilers but the oldest I tried it with is gcc 9.3.1 and I didn't try any non-gcc compilers. I also didn't update the gtk2 codebase (would anybody be using a mix of that and gcc 15?).

### DIFF
--- a/gtkwave3-gtk3/src/gtk23compat.h
+++ b/gtkwave3-gtk3/src/gtk23compat.h
@@ -52,7 +52,7 @@
 #define WAVE_GTK3_GESTURE_ZOOM_IS_1D
 #endif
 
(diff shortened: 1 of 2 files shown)


tbybell:
Paul, these changes compile fine on a Centos7 box I have for running Quartus.  I'll merge the pull request, thanks.

526:/home/bybell/gtkwave/code/gtkwave3> gcc --version
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)

PR #407, opened 2025-01-15
Fixes for tcl9 compatibility

pghmcfc:
1. `CONST` no longer defined in `tcl.h`
 This was a helper for non ANSI-C compilers, replaced by `const` going forwards.

2. ` Tcl_FindExecutable` now returns `const char *` rather than `void`
 See https://www.tcl-lang.org/man/tcl9.0/TclLib/FindExec.html

I only applied the changes in the gtk3 version; would anybody be using gtk2 and tcl9?

### DIFF
--- a/gtkwave3-gtk3/src/tcl_commands.c
+++ b/gtkwave3-gtk3/src/tcl_commands.c
@@ -47,7 +47,7 @@
 
 #if defined(HAVE_LIBTCL)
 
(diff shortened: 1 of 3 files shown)


tbybell (approved):

tbybell:
Paul, thanks for these patches.  I don't see any need to update the gtk2 version. At this point, gtk3 has been out for 14 or so years.  If for some weird reason I ever need a gtk2 w/Tcl 9 version, I'll do the appropriate search and replace.  :-)

PR #444, opened 2025-06-18
Fix parsing of signal names containing colon (:)

### LINKED ISSUE #443: GTKWave does not support signal names with a colon `:`
Some EDA tools will generate VCD files where signal names contain a colon `:`  

Is it possible to support signal names with colons? Thanks!

Image

EPTansuo:
Fixes #443

### DIFF
--- a/lib/libgtkwave/src/gw-vcd-loader.c
+++ b/lib/libgtkwave/src/gw-vcd-loader.c
@@ -787,6 +787,7 @@ static int get_vartoken(GwVcdLoader *self, int match_kw)
 {
     int ch;
     int len = 0;
+    static int in_bracket = 0;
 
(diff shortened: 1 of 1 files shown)


tbybell:
Instead of a static var local to the tokenizer, the in_bracket state probably should be set/cleared in functions such as `vcd_parse_var_regular()` and `vcd_parse_var_evcd()` or elsewhere in case this code eventually becomes re-entrant.  I'll apply a similar fix to LTS for 3.3.125 after some amount of testing that abuses colons in $var names.  It looks like $scope is OK because it's not using vartoken functions.

EPTansuo:
Thank you for your reply. I agree that it would be much better not to use a static variable.

PR #447, opened 2025-06-22
auto reload wave file

DrAtomic:
add a menu checkbox that will enable auto reloading the waveform when the wave file gets updated

### DIFF
--- a/src/menu.c
+++ b/src/menu.c
@@ -2630,10 +2630,7 @@ static void menu_toggle_auto_reload(gpointer null_data, guint callback_action, G
         g_signal_connect(reload_monitor, "changed", G_CALLBACK(file_changed_cb), NULL);
         g_object_unref(file);
     } else {
-        if (reload_monitor) {
-            g_object_unref(reload_monitor);
-            reload_monitor = NULL;
-        }
+        g_clear_object(&reload_monitor);
     }
 }
 
--- a/src/menu.c
+++ b/src/menu.c
@@ -2627,8 +2627,13 @@ static void menu_toggle_auto_reload(gpointer null_data, guint callback_action, G
     if (auto_reload_enabled) {
         GFile *file = g_file_new_for_path(GLOBALS->loaded_file_name);
         reload_monitor = g_file_monitor_file(file, G_FILE_MONITOR_NONE, NULL, NULL);
-        g_signal_connect(reload_monitor, "changed", G_CALLBACK(file_changed_cb), NULL);
-        g_object_unref(file);
+        if (reload_monitor == NULL) {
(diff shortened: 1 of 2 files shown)


src/menu.c
```diff
@@ -2584,21 +2584,67 @@ void menu_new_viewer_tab(gpointer null_data, guint callback_action, GtkWidget *w
...
+    (void)null_data;
+    (void)callback_action;
+    (void)widget;
+
+    static GFileMonitor *reload_monitor;
+    gboolean auto_reload_enabled = gtk_check_menu_item_get_active(GTK_CHECK_MENU_ITEM(menu_wlist[WV_MENU_AFRW]));
+
+    if (auto_reload_enabled) {
+        GFile *file = g_file_new_for_path(GLOBALS->loaded_file_name);
+        reload_monitor = g_file_monitor_file(file, G_FILE_MONITOR_NONE, NULL, NULL);
```
rfuest: `g_file_monitor_file` can fail and return `NULL`. After the UI rewrite errors like this should be shown in the UI, but for now adding a `g_warning` that the file couldn't be monitored is enough. On error the `g_signal_connect` line must not be called and the menu item should be unchecked.

rfuest (changes requested):
Sorry for missing the PR and thanks for keeping it updated.

I think it would also be a good idea to add a command line option to enable file monitoring, so that the user doesn't have to enable each time they run the UI. But that could also be added in a future PR.

src/menu.c
tbybell: Auto reload is a bit harder than it seems.  It's part of the reason why I added the RPC version of reload in LTS in the first place.  You put the RPC invocation in your script to kick off when VCS or whatever finishes, otherwise there are cases where auto reload guesses wrong.

For smaller models, it's OK.  On industrial-sized designs, an undesired spurious reload can be seriously annoying.  It gets tricky with block-based formats with large buffer sizes.  YMMV.

rfuest: That's a good point. The planned feature to allow loading of multiple dump files in a single project will make this problem even harder.

On Linux it would be possible to only reload on `G_FILE_MONITOR_EVENT_CHANGES_DONE_HINT` events, which are triggered when the writing application closes the file. This should at least be safer than trying to reload on any change, but that event doesn't seem to be supported on Windows.

DrAtomic:
I greatly appreciate you taking the time to review this. I have been using this feature for several months and it has been really nice. 

I updated the PR with your requests.

thank you for maintaining this great project!

DrAtomic:
> I think it would also be a good idea to add a command line option to enable file monitoring, so that the user doesn't have to enable each time they run the UI. But that could also be added in a future PR.

I went ahead and added this because I kept forgetting to set the toggle.

PR #456, opened 2025-09-30
Performance improvement when loading big *.ghw files

simonebri:
I've added a simple index cache in ghw_read_cycle_cont (libghw.c) in order to speedup the lookup of non null signals.
The cache is currently freed on ghw_close, but there might be a better place to free it earlier.

### DIFF
--- a/lib/libghw/libghw.c
+++ b/lib/libghw/libghw.c
@@ -2148,6 +2148,12 @@ ghw_close (struct ghw_handler *h)
 
       h->stream = NULL;
     }
+
+  /* Free up the non null signal cache if still in memory */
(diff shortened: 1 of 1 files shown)


rfuest:
Thanks for the PR.

`libghw` is maintained upstream by the GHDL project (https://github.com/ghdl/ghdl/tree/master/ghw). Please open a PR there to let them take a look at your changes first. We can merge your changes into GTKWave when after they have been accepted upstream.

rfuest:
Thanks for letting me know and getting this change merged into upstream. I've applied the changes to the development branch in #470.

PR #470, opened 2025-12-15
Update to latest libghw for improved performance

rfuest:
See #456 for information.

@tbybell: I think this should also be applied to LTS.


tbybell:
b6c7a12..4052649  lts -> lts

Thanks!

PR #472, opened 2025-12-30
Support for Broadway backend instead of X11/XQuartz

nataliakokoromyti:
GTKWave uses GtkPlug/GtkSocket for the -X flag and twinwave embedding, which only work on X11. This PR detects non-X11 backends at runtime and disables these features gracefully instead of crashing.

Changes:

1. main.c: Detect non-X11 display, ignore -X flag with warning
2. twinwave.c: Detect non-X11 display, disable embedded mode
3. Added test to verify -X flag behavior on non-X11 backends

Tested on Broadway and Wayland.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -24,6 +24,7 @@
 #include <windows.h>
 #else
 #ifdef GDK_WINDOWING_X11
+#include <gtk/gtkx.h>
 #include <gdk/gdkx.h>
 #endif
 #endif


src/tests/backend-xid-nonx11.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
```
rfuest: I appreciate that you tried to add a test for this. But it is skipped in CI, which makes it less useful.

Automated tests of the UI is something I want to look into for the GTK 4 port, but at the moment I would suggest to remove the test.

src/main.c
```diff
@@ -1232,6 +1235,24 @@ int main_2(int opt_vcd, int argc, char *argv[])
         }
     }
 
+#ifdef WAVE_USE_XID
+    if (GLOBALS->socket_xid) {
+        gboolean is_x11_display = FALSE;
+#ifdef GDK_WINDOWING_X11
+        {
+            GdkDisplay *display = gdk_display_get_default();
+            if (display && GDK_IS_X11_DISPLAY(display)) {
```
rfuest: `GDK_IS_X11_DISPLAY` already checks for `display != NULL`.
```suggestion
            if (GDK_IS_X11_DISPLAY(display)) {
```

rfuest (changes requested):
Thanks for the PR.

src/tests/backend-xid-nonx11.sh
nataliakokoromyti: i was getting crlf warnings from powershell so i added this as precaution to ensure the test would work on linux CI even when committed from windows. but since the test is not run, .gitattributes can also go.

PR #474, opened 2026-01-04
add libwcp (ADDING WCP SUPPORT PART #1)

nataliakokoromyti:
This PR adds a new standalone library lib/libwcp that implements the WCP protocol core, based on Surfer’s WCP implementation on gitlab.
 
What’s in this PR:

  - wcp_protocol.*: JSON parsing for the full WCP command set + response/event builders.
  - wcp_server.*: small TCP server using GLib/GIO that reads line‑delimited JSON and dispatches to a callback.

Supported commands:
get_item_list, get_item_info, set_item_color, add_variables, add_scope, add_items, add_markers, remove_items, focus_item, clear, set_viewport_to, set_viewport_range, zoom_to_fit, load, reload, shutdown, plus the waveforms_loaded event.

This is the first part of adding wcp support to gtkwave. The integration (handlers + CLI flags) is in a separate PR per the refactor request.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -71,13 +71,7 @@ option(
 option(
     'build_macos_app',
     type: 'boolean',
-    value: 'false',
+    value: false,
     description: 'Build MacOS application bundle',
 )
 
-option(
-    'wcp',
-    type: 'boolean',
-    value: true,
-    description: 'Enable WCP (Waveform Control Protocol) support',
-)
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -3,7 +3,4 @@ subdir('libjrb')
 subdir('liblxt')
 subdir('libvzt')
 subdir('libgtkwave')
-
-if wcp_enabled
-    subdir('libwcp')
-endif
+subdir('libwcp')
--- a/meson.build
+++ b/meson.build
@@ -21,8 +21,6 @@ cc = meson.get_compiler('c')
 add_project_arguments('-D_GNU_SOURCE', language: 'c')
 link_args = []
 
-wcp_enabled = get_option('wcp')
-
 datadir_gtkwave = get_option('datadir') / 'gtkwave3'
 
 version_h = vcs_tag(input: 'src/version.h.in', output: 'version.h')
@@ -31,6 +29,7 @@ version_h = vcs_tag(input: 'src/version.h.in', output: 'version.h')
 
 glib_req = '>=2.72.0'
 glib_req_macro = 'GLIB_VERSION_2_72'
+json_glib_req = '>=1.8.0'
 gtk_req = '>=3.24.0'
 gtk_mac_integration_req = '>=3.0.0'
 gtk4_req = '>=4.6'
@@ -41,6 +40,7 @@ bzip2_req = '>=1.0.0'
 
 glib_dep = dependency('glib-2.0', version: glib_req)
 gobject_dep = dependency('gobject-2.0', version: glib_req)
+gio_dep = dependency('gio-2.0', version: glib_req)
 gtk_dep = dependency('gtk+-3.0', version: gtk_req)
 gtk4_dep = dependency('gtk4', version: gtk4_req)
 gtk_unix_print_dep = dependency(
@@ -53,6 +53,7 @@ gtk_mac_integration_dep = dependency(
     version: gtk_mac_integration_req,
     required: host_machine.system() == 'darwin',
 )
+json_glib_dep = dependency('json-glib-1.0', version: json_glib_req)
 zlib_dep = dependency('zlib', version: zlib_req)
 m_dep = cc.find_library('m', required: false)
 judy_dep = cc.find_library(
@@ -66,20 +67,6 @@ gnu_regex_dep = cc.find_library(
     required: host_machine.system() == 'windows',
 )
 thread_dep = dependency('threads', required: false)
-
-if wcp_enabled
-    gio_dep = dependency('gio-2.0', version: glib_req, required: false)
-    json_glib_dep = dependency('json-glib-1.0', required: false)
-    if not gio_dep.found() or not json_glib_dep.found()
-        wcp_enabled = false
-        message('WCP disabled: missing gio-2.0 or json-glib-1.0')
-    else
-        add_project_arguments(
-            '-DHAVE_WCP',
-            language: 'c',
-        )
-    endif
-endif
 
 bzip2_dep = dependency('bzip2', version: bzip2_req, required: false)
 if not bzip2_dep.found()


lib/libwcp/meson.build
```diff
@@ -0,0 +1,33 @@
...
+    'wcp_server.c',
+]
+
+libwcp_headers = [
+    'wcp_protocol.h',
+    'wcp_server.h',
+]
+
+json_glib_dep = dependency('json-glib-1.0')
+gio_dep = dependency('gio-2.0')
```
rfuest: Dependencies should only be defined in the `meson.build` in the root directory. Both dependencies will also be used by the rest of GTKWave in the future.

lib/libwcp/wcp_protocol.c
```diff
@@ -0,0 +1,751 @@
...
+    if (g_str_equal(cmd_str, "add_markers"))        return WCP_CMD_ADD_MARKERS;
+    if (g_str_equal(cmd_str, "remove_items"))       return WCP_CMD_REMOVE_ITEMS;
+    if (g_str_equal(cmd_str, "focus_item"))         return WCP_CMD_FOCUS_ITEM;
+    if (g_str_equal(cmd_str, "clear"))              return WCP_CMD_CLEAR;
+    if (g_str_equal(cmd_str, "set_viewport_to"))    return WCP_CMD_SET_VIEWPORT_TO;
+    if (g_str_equal(cmd_str, "set_viewport_range")) return WCP_CMD_SET_VIEWPORT_RANGE;
+    if (g_str_equal(cmd_str, "zoom_to_fit"))        return WCP_CMD_ZOOM_TO_FIT;
+    if (g_str_equal(cmd_str, "load"))               return WCP_CMD_LOAD;
+    if (g_str_equal(cmd_str, "reload"))             return WCP_CMD_RELOAD;
+    if (g_str_equal(cmd_str, "shutdown"))           return WCP_CMD_SHUTDOWN;
```
rfuest: The formatting of this block will be changed if you run `clang-format` with the config in the root of the project. If you want to preserve the special format you must add special comments, like in this example: https://github.com/gtkwave/gtkwave/blob/4462c47068a0c0b5c931c54b29df0db8695f8002/lib/libgtkwave/src/gw-color-theme.c#L69-L115

lib/libwcp/wcp_protocol.c
```diff
@@ -0,0 +1,751 @@
...
+        }
+    }
+    json_builder_end_array(builder);
+    
+    json_builder_end_object(builder);
+    
+    return wcp_json_builder_to_string(builder);
+}
+
+gchar* wcp_create_item_list_response(GArray *ids)
```
rfuest: Isn't this the same as `wcp_create_add_items_response_for("get_item_list", ids)`?

lib/libwcp/wcp_protocol.h
```diff
@@ -0,0 +1,167 @@
...
+
+/* Create JSON response messages */
+gchar* wcp_create_greeting(void);
+gchar* wcp_create_ack(void);
+gchar* wcp_create_error(const gchar *error_type, 
+                        const gchar *message,
+                        GPtrArray *arguments);
+gchar* wcp_create_item_list_response(GArray *ids);
+gchar* wcp_create_item_info_response(GPtrArray *items);
+gchar* wcp_create_add_items_response_for(const gchar *command, GArray *ids);
```
rfuest: The naming of some methods doesn't make it clear that they are related to responses (e.g. `wcp_create_error`). I would prefer them to all have "response" in the name. For example:
```suggestion
/* Create JSON response messages */
gchar* wcp_response_greeting(void);
gchar* wcp_response_ack(void);
gchar* wcp_response_error(const gchar *error_type, const gchar *message, GPtrArray *arguments);
// gchar* wcp_create_item_list_response(GArray *ids);  <-- see comment above
gchar* wcp_response_item_info(GPtrArray *items);
gchar* wcp_response_id_list(const gchar *command, GArray *ids);
```
But something like `wcp_create_response_...`, `wcp_build_response...`, or `wcp_response_..._new` would also be OK, if you prefer that.

lib/libwcp/wcp_protocol.h
```diff
@@ -0,0 +1,167 @@
...
+
+/* Create JSON event messages */
+gchar* wcp_create_waveforms_loaded_event(const gchar *source);
```
rfuest: This function should also be renamed to match the response methods, just with `event` in the name instead.

lib/libwcp/wcp_server.c
```diff
@@ -0,0 +1,308 @@
...
+    if (server->client_connection) {
+        g_io_stream_close(G_IO_STREAM(server->client_connection), NULL, NULL);
+        g_object_unref(server->client_connection);
+        server->client_connection = NULL;
+    }
+    
+    if (server->data_input) {
+        g_object_unref(server->data_input);
+        server->data_input = NULL;
+    }
```
rfuest: Streams are automatically closed when the object is freed.
```suggestion
    g_clear_object(&server->client_connection);
    g_clear_object(&server->data_input);
```

meson_options.txt
```diff
@@ -74,3 +74,17 @@ option(
...
+    value: false,
+    description: 'Enable WCP (Waveform Control Protocol) support',
+)
+
+option(
+    'wcp_default_port',
+    type: 'integer',
+    value: 8765,
+    description: 'Default port for WCP server',
+)
```
rfuest: I don't think this needs to be a build time option. The WCP library can just always take a port in the server constructor and the GTKWave code can later add a default value and the possibility to override it with a command line argument or configuration file option.

lib/libwcp/wcp_server.c
```diff
@@ -0,0 +1,308 @@
...
+void wcp_server_free(WcpServer *server)
+{
+    if (!server) return;
+    
+    wcp_server_stop(server);
+    g_free(server);
+}
+
+gboolean wcp_server_send(WcpServer *server, gchar *message)
+{
```
rfuest:
```suggestion
{
    g_return_val_if_fail(server != NULL, FALSE);
    g_return_val_if_fail(message != NULL, FALSE);    
```

lib/libwcp/wcp_server.c
```diff
@@ -0,0 +1,308 @@
...
+    
+    gchar *event = wcp_create_waveforms_loaded_event(source);
+    wcp_server_send(server, event);
+}
+
+gboolean wcp_server_initiate(WcpServer *server,
+                             const gchar *host,
+                             guint16 port,
+                             GError **error)
+{
```
rfuest:
```suggestion
{
    g_return_val_if_fail(server != NULL, FALSE);
    g_return_val_if_fail(host != NULL, FALSE);
    g_return_val_if_fail(port != 0, FALSE);
    g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```

lib/libwcp/meson.build
```diff
@@ -0,0 +1,33 @@
+libwcp_sources = [
+    'wcp_protocol.c',
+    'wcp_server.c',
+]
+
+libwcp_headers = [
+    'wcp_protocol.h',
+    'wcp_server.h',
+]
+
```
rfuest:
```suggestion
```

lib/libwcp/wcp_protocol.c
```diff
@@ -0,0 +1,700 @@
...
+static GArray* parse_id_array(JsonArray *arr, GError **error)
+{
+    GArray *ids = g_array_new(FALSE, FALSE, sizeof(WcpDisplayedItemRef));
+    size_t len = (size_t)json_array_get_length(arr);
+    
+    for (size_t i = 0; i < len; i++) {
+        JsonNode *node = json_array_get_element(arr, i);
+        if (!JSON_NODE_HOLDS_VALUE(node)) {
+            g_set_error(error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA,
+                        "ids[%u] must be a number", i);
```
rfuest: This line causes a compiler warning, because the type of `i` has changed. This also applies to other `g_set_error` statements in this function and the two functions below.
```suggestion
            g_set_error(error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA,
                        "ids[%zu] must be a number", i);
```

rfuest (changes requested):
When I tried to build the branch locally the build failed because of a missing `case` statement in `wcp_protocol.c`. CI hadn't caught the issue because `json-glib` wasn't installed. I've pushed a commit to install the necessary dependency, which makes CI actually build the WCP code.

PR #475, opened 2026-01-12
wire wcp (ADDING WCP SUPPORT PART #2)

nataliakokoromyti:
This PR wires the merged libwcp (PR #474) into GTKWave. It adds the GTKWave‑specific command handlers, CLI flag, and event wiring so external tools can control GTKWave
  over WCP.

  What it does

  - adds --wcp-port CLI flag and starts the WCP server in src/main.c
  - emits waveforms_loaded after load/reload
  - links GTKWave against libwcp in the build system

  Manual test

  1. Start GTKWave:
     ./build/src/gtkwave --wcp-port=8765 ./examples/des.fst
  2. In another terminal connect:
     nc localhost 8765
  3. Send commands like add_items, get_item_list, set_viewport_range, etc.

### DIFF
--- a/src/wcp_gtkwave.c
+++ b/src/wcp_gtkwave.c
@@ -399,42 +399,6 @@ static char* handle_add_variables(WcpServer *server, WcpCommand *cmd)
     return response;
 }
 
-static char* handle_add_scope(WcpServer *server, WcpCommand *cmd)
-{
-    (void)server;
-    
-    if (!GLOBALS->dump_file || GLOBALS->loaded_file_type == MISSING_FILE) {
-        return wcp_response_error("no_waveform", "No waveform loaded", NULL);
-    }
-
-    GPtrArray *added_ids = g_ptr_array_new_with_free_func(g_free);
-    
-    GwTreeNode *scope = wcp_find_scope_node(cmd->data.add_scope.scope);
-    if (!scope) {
-        g_ptr_array_free(added_ids, TRUE);
-        return wcp_response_error("invalid_scope", "Scope not found", NULL);
-    }
-
-    GPtrArray *symbols = g_ptr_array_new();
-    wcp_collect_scope_symbols(scope->child, cmd->data.add_scope.recursive, symbols);
-
-    GHashTable *added_vec_roots = g_hash_table_new(g_direct_hash, g_direct_equal);
-    for (unsigned int i = 0; i < symbols->len; i++) {
-        GwSymbol *sym = g_ptr_array_index(symbols, i);
-        wcp_add_symbol(sym, added_vec_roots, added_ids);
-    }
-    g_hash_table_destroy(added_vec_roots);
-    g_ptr_array_free(symbols, TRUE);
-    
-    GLOBALS->signalwindow_width_dirty = 1;
-    MaxSignalLength();
-    redraw_signals_and_waves();
-    
-    char *response = wcp_response_id_list("add_scope", added_ids);
-    g_ptr_array_free(added_ids, TRUE);
-    return response;
-}
-
 static char* handle_add_markers(WcpServer *server, WcpCommand *cmd)
 {
     (void)server;
@@ -712,22 +676,6 @@ static char* handle_zoom_to_fit(WcpServer *server, WcpCommand *cmd)
     return wcp_response_ack();
 }
 
-static gboolean wcp_stop_server_idle(gpointer data)
-{
-    wcp_server_stop((WcpServer *)data);
-    return G_SOURCE_REMOVE;
-}
-
-static char* handle_shutdown(WcpServer *server, WcpCommand *cmd)
-{
-    (void)cmd;
-
-    /* Stop after the ack has been sent. */
-    g_idle_add(wcp_stop_server_idle, server);
-    
-    return wcp_response_ack();
-}
-
 /* ============================================================================
  * Main Command Dispatcher
  * ============================================================================ */
@@ -748,9 +696,6 @@ static char* wcp_command_handler(WcpServer *server, WcpCommand *cmd, gpointer us
             
         case WCP_CMD_ADD_VARIABLES:
             return handle_add_variables(server, cmd);
-            
-        case WCP_CMD_ADD_SCOPE:
-            return handle_add_scope(server, cmd);
             
         case WCP_CMD_ADD_ITEMS:
             return handle_add_items(server, cmd);
@@ -781,9 +726,6 @@ static char* wcp_command_handler(WcpServer *server, WcpCommand *cmd, gpointer us
             
         case WCP_CMD_RELOAD:
             return handle_reload(server, cmd);
-            
-        case WCP_CMD_SHUTDOWN:
-            return handle_shutdown(server, cmd);
             
         default:
             return wcp_response_error("unknown_command", 
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,6 +4,7 @@ Context support is Copyright (C) 2007-2024 Kermin Elliott Fleming.
 Trace group support is  Copyright (C) 2009-2024 Donald Baltus.
 GHW and additional GUI support is Copyright (C) 2005-2024 Tristan Gingold.
 Analog support is Copyright (C) 2005-2024 Thomas Sailer.
+WCP support is Copyright (C) 2025-2026 Natalia Kokoromyti.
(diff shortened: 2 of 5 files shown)


src/main.c
```diff
@@ -31,6 +31,10 @@
 
 #include "wave_locale.h"
 
+#ifdef HAVE_WCP
```
rfuest: It's not necessary to check for `HAVE_WCP`, because it's always defined.

src/main.c
```diff
@@ -272,6 +276,13 @@ static void print_help(char *nam)
 #else
 #define RPC_GETOPT
 #define RPC_GETOPT3
+#endif
+
+#ifdef HAVE_WCP
+#define WCP_GETOPT \
+    "  --wcp-port=PORT            start WCP server on PORT (0 uses 8765)\n"
```
rfuest: I prefer to split this into two options, one to enable the server and another to specify the port:
```
--wcp    enable WCP server
--wcp-port=PORT   set WCP port (defaults to 8765)
```

src/wcp_gtkwave.c
```diff
@@ -0,0 +1,867 @@
...
+#include "zoombuttons.h"
+
+/* Global WCP server instance */
+WcpServer *g_wcp_server = NULL;
+
+#define WCP_ITEM_MARKER_FLAG (1ull << 62)
+
+static GHashTable *wcp_trace_to_id = NULL;
+static GHashTable *wcp_id_to_trace = NULL;
+static uint64_t wcp_next_trace_id = 1;
```
rfuest: To reduce the number of global variables these variables should be declared in a struct:
```suggestion
typedef struct {
    GHashTable *trace_to_id;
    GHashTable *id_to_trace;
    uint64_t next_id;
} WcpTraceMap;

static WcpTraceMap *wcp_trace_map = NULL;
```

src/wcp_gtkwave.c
```diff
@@ -0,0 +1,867 @@
...
+    uint64_t *trace_id = g_new(uint64_t, 1);
+    *trace_id = id;
+    g_hash_table_insert(wcp_trace_to_id, t, trace_id);
+    uint64_t *id_key = g_new(uint64_t, 1);
+    *id_key = id;
+    g_hash_table_insert(wcp_id_to_trace, id_key, t);
+    return id;
+}
+
+static void wcp_remove_trace_id(GwTrace *t)
```
rfuest:
```suggestion
static void wcp_trace_map_remove_trace_id(WcpTraceMap *self, GwTrace *t)
```

src/wcp_gtkwave.c
```diff
@@ -0,0 +1,867 @@
...
+{
+    if (!wcp_id_to_trace) {
+        return NULL;
+    }
+    int64_t key = (int64_t)id;
+    GwTrace *t = g_hash_table_lookup(wcp_id_to_trace, &key);
+    if (!t) {
+        return NULL;
+    }
+    return t;
```
rfuest: This should do the same:
```suggestion
    return t;
```

src/wcp_gtkwave.c
```diff
@@ -0,0 +1,867 @@
...
+
+static int wcp_parse_color(const char *color)
+{
+    if (!color || !*color) {
+        return -1;
+    }
+
+    if (g_ascii_isdigit(color[0])) {
+        char *end = NULL;
+        long val = strtol(color, &end, 10);
```
rfuest:
```suggestion
        uint64_t val = g_ascii_strtoll(color, &end, 10);
```

src/wcp_gtkwave.c
```diff
@@ -0,0 +1,867 @@
...
+                    gw_marker_set_alias(marker, NULL);
+                }
+                continue;
+            }
+
+            GwTrace *t = wcp_lookup_trace(ref->id);
+            if (t) {
+                wcp_remove_trace_id(t);
+                RemoveTrace(t, 1);
+            }
```
rfuest: Using an `else` instead of the `continue` is easier to read:
```suggestion
            } else {
                GwTrace *t = wcp_lookup_trace(ref->id);
                if (t) {
                    wcp_remove_trace_id(t);
                    RemoveTrace(t, 1);
                }
            }
```

src/wcp_gtkwave.h
```diff
@@ -0,0 +1,46 @@
...
+
+/* ============================================================================
+ * Initialization
+ * ============================================================================ */
+
+/**
+ * Initialize WCP support for GTKWave
+ * Call this from main() after GTKWave is initialized
+ * 
+ * @param port Port to listen on (0 for default 8765)
```
rfuest: `port` is no longer allowed to be `0`.

src/wcp_gtkwave.c
```diff
@@ -0,0 +1,867 @@
...
+#include "lx2.h"
+#include "signal_list.h"
+#include "symbol.h"
+#include "wavewindow.h"
+#include "zoombuttons.h"
+
+/* Global WCP server instance */
+WcpServer *g_wcp_server = NULL;
+
+#define WCP_ITEM_MARKER_FLAG (1ull << 62)
```
rfuest: Using such a large value as a flag can cause issues. All numeric values in a JSON file are doubles, which only have 53 significant bits (see last paragraph of https://datatracker.ietf.org/doc/html/rfc8259#section-6). While some JSON libraries can also handle larger integers, this is not portable.

Alternatives would be to use a tagged union to store all items in a single array or to use string IDs like `T42` or `M2` (which is also allowed according to https://gitlab.com/waveform-control-protocol/wcp/-/tree/main?ref_type=heads#displayeditemref).

rfuest (changes requested):
I haven't had time to try it yet, but here is a first review pass. Most of it looks good, but the way item IDs are generated will need to be modified (see comment about `WCP_ITEM_MARKER_FLAG`).

src/main.c
```diff
@@ -272,6 +274,14 @@ static void print_help(char *nam)
 #else
 #define RPC_GETOPT
 #define RPC_GETOPT3
+#endif
+
+#ifdef HAVE_WCP
+#define WCP_GETOPT \
+    "  --wcp                     enable WCP server\n" \
+    "  --wcp-port=PORT            set WCP port (defaults to 8765)\n"
```
rfuest: At the moment the WCP server listens for connections on all interfaces, which means that it is openly accessible to anyone on the same network. For security reasons the default should be to only listen for connections from localhost. An additional flag could then be used to also allow remote connections.

src/main.c
```diff
@@ -977,6 +991,18 @@ int main_2(int opt_vcd, int argc, char *argv[])
...
+                    if (!strcmp(long_options[option_index].name, "wcp")) {
+                        wcp_enable = 1;
+                    } else if (!strcmp(long_options[option_index].name, "wcp-port")) {
+                        wcp_port = atoi(optarg);
+                        wcp_enable = 1;
+                    } else if (!strcmp(long_options[option_index].name, "wcp-remote")) {
+                        wcp_allow_remote = 1;
+                        wcp_enable = 1;
+                    }
+                    break;
```
rfuest: In my opinion the other options shouldn't imply `--wcp`.
```suggestion
                case 0:
                    if (!strcmp(long_options[option_index].name, "wcp")) {
                        wcp_enable = 1;
                    } else if (!strcmp(long_options[option_index].name, "wcp-port")) {
                        wcp_port = atoi(optarg);
                    } else if (!strcmp(long_options[option_index].name, "wcp-remote")) {
                        wcp_allow_remote = 1;
                    }
                    break;
```

src/wcp_gtkwave.c
```diff
@@ -0,0 +1,852 @@
...
+{
+    WcpTraceMap *self = g_new0(WcpTraceMap, 1);
+    if (!self) {
+        if (!wcp_trace_map_fallback_inited) {
+            wcp_item_map_init(&wcp_trace_map_fallback.traces, 'T');
+            wcp_item_map_init(&wcp_trace_map_fallback.markers, 'M');
+            wcp_trace_map_fallback_inited = TRUE;
+        }
+        return &wcp_trace_map_fallback;
+    }
```
rfuest: I don't understand this code and what `wcp_trace_map_fallback` is used for. `self` is always non null after `g_new0` is called, so that the code inside the if statement is never called.

PR #480, opened 2026-01-29
docs: add dependency on libjson-glib to build instructions

supervacuus:
The recent changes in https://github.com/gtkwave/gtkwave/pull/474 make `libjson-glib` a required dependency for the default build. This is reflected in the build configuration but not yet in the documentation.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,6 +4,7 @@ Context support is Copyright (C) 2007-2024 Kermin Elliott Fleming.
 Trace group support is  Copyright (C) 2009-2024 Donald Baltus.
 GHW and additional GUI support is Copyright (C) 2005-2024 Tristan Gingold.
 Analog support is Copyright (C) 2005-2024 Thomas Sailer.
+WCP support is Copyright (C) 2025-2026 Natalia Kokoromyti.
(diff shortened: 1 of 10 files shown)


docs/install/unix_linux_4.md
```diff
@@ -13,14 +13,15 @@ Debian, Ubuntu:
 sudo apt install build-essential meson gperf libbz2-dev
 sudo apt install flex desktop-file-utils libgtk-3-dev git
 sudo apt install libgirepository1.0-dev
+sudo apt install libjson-glib-dev
 sudo apt install libjudy-dev # optional
```
rfuest: `libgtk-4-dev` was also missing in this list. I've tried to sort the packages a bit, by tools first and then libraries. `libjudy` won't be required anymore soon and is already optional, so it's safe to remove it:

```suggestion
sudo apt install \
    build-essential \
    git \
    meson \
    gperf \
    flex \
    desktop-file-utils \
    libbz2-dev \
    libgtk-3-dev \
    libgtk-4-dev \    
    libgirepository1.0-dev \
    libjson-glib-dev
```

PR #481, opened 2026-01-31
Automatically turn fixed-point shift on when the shift is specified

Kwekker:
There is code in `menu_dataformat_fpshift_specify_cleanup` that automatically turns fixed-point shift _on_, which is very useful behavior in my opinion.
I removed a line that automatically turns fixed-point shift _off_ again after this cleanup callback is run.

This has been bugging me for a while.

### DIFF
--- a/src/menu.c
+++ b/src/menu.c
@@ -4343,8 +4343,6 @@ void menu_dataformat_fpshift_specify(gpointer null_data, guint callback_action, 
              NULL,
              128,
              G_CALLBACK(menu_dataformat_fpshift_specify_cleanup));
-
-    dataformat(~(TR_FPDECSHIFT), 0);
 }
 
 void menu_dataformat_invert_on(gpointer null_data, guint callback_action, GtkWidget *widget)


tbybell:
aae8386..ee041a8  lts -> lts

rfuest:
@tbybell: The fix you made in LTS is subtly different from this one. `menu_dataformat_fpshift_specify_cleanup` does disable fpshift when the user inputs an invalid shift value. By unconditionally enabling the fpshift flag in `menu_dataformat_fpshift_specify` the flags could get modified without a forced redraw of the traces. Only setting the flags in the cleanup function should be enough and the function contains the necessary redraw (it even unnecessarily redraws twice in some cases).

rfuest:
Forget what I said about the traces not updated, because I forgot that `dataformat` also updates the traces. But the part about the fpshift flag being set unconditionally and overriding the conditional logic in `menu_dataformat_fpshift_specify_cleanup` still applies.

tbybell:
Thanks for spotting that.  There already is a form of dataformat() augmented with shamt in the cleanup callback for entrybox so the dataformat() in the actual menu item code itself can be removed.

ee041a8..35938cd  lts -> lts

Issue #486, opened 2026-02-16
[Feature request] Please add daily builds of gtkwave


zapta:
Please add github workflow that builds and publishes daily builds of gtkwave for macos/win/linux. This will help users that want to test or use the latest features and fixes.

If it's ok with the gtkwave team, I have experience with similar workflows and can help. For example, this is a workflow I wrote for daily builds of Apio.  https://github.com/FPGAwars/apio/blob/main/.github/workflows/build-pre-release.yaml

rfuest:
It's pretty outdated and it might be better to just ignore it. There is, for example, no need to support 32 bit Windows anymore. Perhaps there is something useful in there when it comes to packaging later, but to get started it should be easier to base a `build-windows` job in `ci.yml` on the build jobs for the other OSes.

The old build used MSYS2's MINGW64 environment, but the new builds should use the UCRT64 one, which is also suggested by GTK: https://www.gtk.org/docs/installations/windows/#using-gtk-from-msys2-packages.

PR #487, opened 2026-02-21
CI job now exports an gtkwave macos arm64 app

zapta:
This PR is the first installment toward https://github.com/gtkwave/gtkwave/issues/486. It extends to CI workflow to build and export as an artifact and stand alone macos arm64 .app with zero brew prerequisites.

Sample artifacts can be found here https://github.com/zapta/gtkwave/actions/workflows/ci.yml

More details in the commit itself.

### DIFF
--- a/macos/launcher.sh
+++ b/macos/launcher.sh
@@ -8,8 +8,4 @@ RESOURCES="$CONTENTS/Resources"
 
 export DYLD_LIBRARY_PATH="$RESOURCES/lib:${DYLD_LIBRARY_PATH:-}"
 
-# TODO: Find a cleaner solution to avoid the warnings.
-# For now we filter out these warnings and empty lines that
-# tend to appear between them.
-
-exec "$RESOURCES/bin/gtkwave" "$@" 2>&1 | grep -v -E '^$|GdkPixbuf-WARNING.*Error loading XPM image loader'
+exec "$RESOURCES/bin/gtkwave" "$@"


.github/workflows/ci.yml
```diff
@@ -27,27 +39,65 @@ jobs:
...
+        run: |
+          sh ./macos/build.sh
+
+          # Export the .app dir location.
+          echo "APP_DIR=${HOME}/Applications/GTKWave.app" >> "$GITHUB_ENV"
+
+      - name: Archive Applications/GTKWave.app
+        run: |
+          # Create the archive under ./build
+          # Per Grok it's safer to use here ditto than zip.
```
rfuest: I don't like the "AI said so" reason for using ditto instead of zip:

```suggestion
          # Use ditto instead of zip to make sure resources forks are preserved in the zip archive.
```

rfuest (changes requested):
I wasn't able to test the build artifact because I don't have access to a macOS machine, but the changes look good to me.

I've added a change to the master branch to replace the deprecated XPM icons with PNGs, which should fix the XPM warnings. Please rebase the branch to see if it removes the warnings that you had to filter out in `launch.sh` and remove the filter.

rfuest:
I think it might be caused by this line, which is missing in the master branch for some reason: https://github.com/gtkwave/gtkwave/blob/0a9dbdb0e6274e968ea02b3d08610918d71f8175/gtkwave3-gtk3/src/main.c#L2180

zapta:
I can trying adding it once this PR is merged, otherwise my attempts will leak to this PR (I build only on github, not locally). 

It also has a related initialization here:
https://github.com/gtkwave/gtkwave/blob/0a9dbdb0e6274e968ea02b3d08610918d71f8175/gtkwave3-gtk3/src/main.c#L2139

@rfuest, what is the status of this PR, anything I should do here?

zapta:
... and here is a similar invocation (not via Apio yet) of the new GTKWave build, all seems to work ok.

```
$ ~/Downloads/GTKWave.app/Contents/MacOS/gtkwave "--rcvar=splash_disable on" "--rcvar=do_initial_zoom_fit 1" "--rcvar=do_initial_zoom_fit 0" _build/tang20k/main_tb.vcd main_tb.gtkw

GTKWave Analyzer v4.0.0-prealpha (1606356) (w)1999-2022 BSI

RCVAR   | 'splash_disable on' NOT FOUND
RCVAR   | 'do_initial_zoom_fit 1' FOUND
RCVAR   | 'do_initial_zoom_fit 0' FOUND
[0] start time.
[40005000] end time.
WM Destroy
```

zapta:
... I also tested launching the new gtkwave from the command line without the .app wrapper dir, similar to how oss-cad-suite and apio uses now it it worked just fine which is very useful for embedded gtkwave use cases.

```
$ mv GTKWave.app/Contents gtkwave
~/Downloads$ ./gtkwave/MacOS/gtkwave 

GTKWave Analyzer v4.0.0-prealpha (1606356) (w)1999-2022 BSI

GTKWAVE | Use the -h, --help command line flags to display help.
WM Destroy
```

PR #490, opened 2026-02-22
Remove .zip suffix from CI artifact name.

zapta:
This eliminates the double suffix '.zip.zip' since Github adds its own zip layer.

No change in the artifact's content.


rfuest (approved):
LGTM

PR #508, opened 2026-04-12
Respect selected signals during recurse import

mehdikazemi2020:
This fixes two related issues in the SST signal-selection flow.

What changed:
- Recurse Import now respects the selected rows in the lower signal list when a signal selection exists.
- Right-clicking in the lower signal list no longer destroys an existing multi-selection before opening the SST popup menu.

Previous behavior:
- Recurse Import ignored the lower-pane selection and imported the entire selected scope.
- Right-clicking after selecting multiple signals collapsed the selection before the import action ran.

New behavior:
- If one or more rows are selected in the lower signal list, Recurse Import operates on those selected rows only.
- If no lower-pane selection exists, Recurse Import keeps the existing scope-wide behavior.
- Right-click on an already selected row preserves the current selection and opens the popup.
- Right-click on an unselected row selects that row and opens the popup.
- Right-click on empty space preserves the existing selection and opens the popup when a selection exists.

Files changed:
- `src/treesearch.c`

Notes:
- This was motivated by interactive SST usage where users select a subset of signals in the lower pane, then use the context menu to import only that subset.

### DIFF
--- a/src/treesearch.c
+++ b/src/treesearch.c
@@ -1787,9 +1787,90 @@ top:
     }
 }
 
+struct recurse_selected_counter
+{
+    gint facilities;
+};
+
+static void recurse_count_selected(GtkTreeModel *model,
+                                   GtkTreePath *path,
+                                   GtkTreeIter *iter,
+                                   gpointer data)
+{
+    (void)path;
+
+    struct recurse_selected_counter *counter = data;
+    GwTreeNode *sel = NULL;
+
+    gtk_tree_model_get(model, iter, TREE_COLUMN, &sel, -1);
+    if (sel == NULL) {
+        return;
+    }
+
+    gint low = fetchlow(sel)->t_which;
+    gint high = fetchhigh(sel)->t_which;
+
+    if ((low < 0) || (high < low)) {
+        return;
+    }
+
+    counter->facilities += high - low + 1;
+}
+
+static enum sst_cb_action recurse_action_to_sst_action(guint callback_action)
+{
+    if (callback_action == WV_RECURSE_INSERT) {
+        return SST_ACTION_INSERT;
+    }
+
+    if (callback_action == WV_RECURSE_REPLACE) {
+        return SST_ACTION_REPLACE;
+    }
+
+    return SST_ACTION_APPEND;
+}
+
 void recurse_import(GtkWidget *widget, guint callback_action)
 {
+    gint fz = 0;
+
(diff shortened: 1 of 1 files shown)


src/treesearch.c
```diff
@@ -1123,7 +1123,37 @@ static gboolean view_selection_func(GtkTreeSelection *selection,
...
+            gtk_tree_path_free(path);
+            do_sst_popup_menu(widget, event);
+            return TRUE;
+        }
+
+        if (gtk_tree_selection_count_selected_rows(selection) > 0) {
+            do_sst_popup_menu(widget, event);
+            return TRUE;
+        }
+    }
```
rfuest, line 1156: I think this should be equivalent and a bit shorter:
```suggestion
    if (event->button == 3 && event->type == GDK_BUTTON_PRESS) {
        GtkTreeView *view = GTK_TREE_VIEW(widget);
        GtkTreeSelection *selection = gtk_tree_view_get_selection(view);
        GtkTreePath *path = NULL;

        if (gtk_tree_view_get_path_at_pos(view, event->x, event->y, &path, NULL, NULL, NULL)) {
            if (!gtk_tree_selection_path_is_selected(selection, path)) {
                gtk_tree_selection_unselect_all(selection);
                gtk_tree_selection_select_path(selection, path);
            }

            gtk_tree_path_free(path);
        }

        if (gtk_tree_selection_count_selected_rows(selection) > 0) {
            do_sst_popup_menu(widget, event);
            return TRUE;
        }
    }
```

rfuest (changes requested):
Thanks for the PR. The changed right click behavior in the bottom pane is a good idea. I've implemented the same for the signal list (where the waveforms are displayed) some time ago. In a later PR the same change could also be applied to double clicking in the bottom pane.

I don't think that recursive import function should work differently when signals in the bottom pane are selected. A user might get confused about why the behavior of clicking in the top pane changes if they have a signal selected in the bottom pane.
A better solution, in my opinion, would be to use two different context menus for the top and bottom panes. The context menu in the top pane would remain unchanged and contain the recursive import option and the bottom pane would contain the non recursive variant.

Note that most of the UI code will be replaced as part of the GTK 4 port in the future. `GtkTreeView`, for example, is deprecated and won't be used anymore. Your PR is still appreciated, because it helps me identify a usability issue and it is an improvement for users of the nightly builds. I feel I should mention this, because I'm struggling myself to find a balance between putting effort in fixing issues in the current UI and waiting for the GTK 4 port, which might not reuse any of these changes.

PR #513, opened 2026-06-20
vcd: recover EVCD port direction and drive strength

wm2015email:
## Summary

GTKWave already loads Extended VCD (`$dumpports`) files, but only by *collapsing* every value change to four-state VCD at parse time — the 9-state EVCD alphabet was flattened to `0/1/x/z` and **both drive-strength tokens were discarded**, so every EVCD port was reported as `GW_VAR_DIR_IMPLICIT`. This PR stops throwing that information away.

## Changes

- **New `gw-evcd.[ch]`** — a single source of truth for EVCD state-char semantics (IEEE 1364-2005 §18.4.3): char→four-state, char→direction, char→human description, and strength-code→name. The loader's `evcd_strcpy()` and the standalone converters now reference it instead of each keeping a private table.
- **Port direction inference** — each port's direction is inferred from the class of state characters it exhibits (input `D U N Z d u`, output `L H X T l h`, unknown-direction `0 1 ? F A a B b C c f`), honoring the `in`/`out`/`inout` keyword form when a file uses it. The result is propagated to `GwNode` and surfaced through `gw_dump_file_has_nonimplicit_directions()`.
- **Drive strength** — the two strength components (0–7) that used to be discarded are captured and propagated to `GwNode`.
- **GUI** — the signal hover/value popup (`gw-mouseover`) now shows the EVCD direction and drive strength for `$dumpports` ports, for both the waveform and the signal list, e.g. `| EVCD: out, strong`.
- **Docs** — an Extended VCD section added to `docs/intro/formats.md`.

The waveform is still drawn in GTKWave's native nine-state alphabet (`0/1/x/z/h/u/w/l`), so the *exact* EVCD conflict character is folded into the displayed value; the recovered direction and strength make the EVCD intent visible without changing the renderer.

## Tests

All 46 `libgtkwave` tests pass, including:
- new `test-gw-evcd` unit test (state-char/direction/strength mappings),
- a programmatic direction + strength check in `test-gw-vcd-loader`,
- a richer `evcd_directions.vcd` dump fixture covering input/output/inout/vector ports, varied strengths, `$dumpportsoff/on`, and `$vcdclose`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tbybell:
```
        if (v->evcd_strength_seen) {
            n->evcd_strength_valid = 1;
            n->evcd_strength0 = v->evcd_strength0 & 0x7;
            n->evcd_strength1 = v->evcd_strength1 & 0x7;
        }
```

@rfuest 
This doesn't look to me like what is proposed in the push works.  Strengths are encountered in the EVCD file attached to a given time during value change parsing so they must be treated as value changes, not as a part of the Node/variable information because subsequent values--which are permitted to be different--encountered at the same port number will overwrite the previous value.  (This gets quite fun once the vlist recoder is a part of the processing...)  If I'm missing something you changed for devel, let me know, but I'm leery of these voluminous dumps of AI-generated code that appear thorough, yet lack knowledge of a program's architecture.  I'm banning all such PRs from LTS unless it's from someone with whom I have a historical working relationship.

I'll refrain from editorializing :-) much further on this.

rfuest:
@tbybell: Thanks for taking a look at the PR, seeing the size and AI generated nature of the PR put it way down on my review pile.

@wm2015email: I won't be merging this PR at the moment. The VCD loader/recoder has undergone a lot more refactoring, which isn't public in the `master` branch yet, and your PR would conflict with these changes. I've never used EVCD myself, so I'm not sure how useful guessing the direction based on the transitions would be. As Tony mentioned, proper EVCD support with per transition data would require much more extensive changes.

